PR #1, opened 2025-01-22
Initial start of model framework and anvil

hmacdope:
Initial start of model framework


hmacdope:
Going to merge here so work can be distributed nicely.

PR #46, opened 2025-02-14
Cross validation prototype

hmacdope:
Prototype of cross validation workflow.


hmacdope:
Some things to fix but this should unblock @dwwest to work on comparisons

Issue #27, opened 2025-02-06
Abstract away call to `model_pkl` to a serialisation / deserialisation function that is implementation agnostic


hmacdope:
Models need a wrapper that serializes and deserializes the model that is independent of the model typing

For example we currently use the following line in `AnvilWorkflow.run()`

```python
        self.model.to_model_json_and_pkl(
            output_dir / "model.json", output_dir / "model.pkl"
        )
```
This needs to become an abstract `serialize()` function with a matching `deserialize`.

These can call model specific versions under the hood if needed, mostly just needs a wrapper for API compliance.

hmacdope:
@dwwest would you be able to tackle this one?

PR #55, opened 2025-02-24
Added serialize/deserialize abstract functions for #27

dwwest:
## Description
We are replacing pickle-specific serialization functions with base class abstract methods "serialize" and "deserialize" to stay API compliant


hmacdope (approved):
LGTM! pending CI passing

Issue #53, opened 2025-02-20
Add ability dimensions of CV split to `cross_validation_metrics.json`


hmacdope:
We are recording the cross validation metrics, but not the shape of the resulting CV grid.

This is important as there are many ways a repeated CV grid could be done resulting in the same shape. ie a set of metrics with 12 entries could have come from a 3x4, 4x3, 2x6 or 1x12 CV setup (illustrative). 

We should add an entry to the resulting `json` file that records the dimensions. 

I am thinking of an entry that just contains the array shape tuple. For example:

```json
shape: [ 5, 5 ]

```

PR #67, opened 2025-03-05
Added cv shape data to cv output json for #53

dwwest:
Needed to save CV dimensions to output json


dwwest:
Not sure if this was supposed to be a one-line change, @hmacdope let me know?

PR #75, opened 2025-03-10
Add container build for ghcr.io

### LINKED ISSUE #74: Add container for production use of repo and dependencies
For RT inference and also for general use we should provide a docker container with repo and dependencies.

hmacdope:
Fixes #74 

Adds a container that contains repo and all dependencies so that we can build RT inference stack.


hmacdope:
Will just merge straight up so that build comes from main.

Issue #76, opened 2025-03-12
Make PDF report from comparison workflows. 


hmacdope:
As per discussion @dwwest we will move PDF generation to new PR.

_Originally posted by @hmacdope in https://github.com/OpenADMET/openadmet_models/issues/63#issuecomment-2716171241_

PR #85, opened 2025-03-14
Add pdf reporting to posthoc analysis for #76

dwwest:
Add pdf reporting to posthoc analysis for #76


dwwest:
I need to change use of name "save_dir" to "output_dir" to be consistent with other modules

hmacdope (changes requested):
Add `statsmodels` to the correct file in `devtools/conda_envs/` and fix sig fig

Issue #42, opened 2025-02-13
Add `LGBMClassifierModel` wrapping `lgbm` classifier


hmacdope:
Wrapper around `LGBMClassifier` for model training.

hmacdope:
@smcolby would you be able to tackle this?

hmacdope:
Will involve implementing the relevant featurizers also

Issue #41, opened 2025-02-13
Implement base classification metrics


hmacdope:
Implement our base classification metrics as a standalone function and as called in an `Evaluator`

- AUPR
- AUROC (?)
- Class statistics
- Cross entropies and related metrics

dwwest:
@hmacdope should we wrap this into the main PostHoc classification since I'm adding a binary classification piece in to treat regression models as classifiers?  Maybe I can take this issue once we have a few models trained?  Or do we want this in eval instead of posthoc?

hmacdope:
We should do in both places I think. The regression -> classifier is more specialised based on a post-hoc threshold whereas the classification metrics themselves will need to be implemented regardless.

hmacdope:
Done by @smcolby

PR #93, opened 2025-03-25
Implement classification infrastructure in Anvil

### LINKED ISSUE #40: Implement first classification models
We should begin to explore implementing our first classification models and metrics. 

These will likely be LGBMClassifiers with associated classification metrics

As part of this we will likely need to introduce a system for checking compatibility between `Anvil` components designed for regression and those designed for classification. This will need to be well thought out as may become fairly clunky otherwise.

smcolby:
## Description
Anvil is currently able to perform regression tasks, but not classification tasks. This PR adds LGBM-based classification functionality, classifier evaluation metrics, and adds each to the registry.

Resolves #40 #41 #42


hmacdope (commented):
TODOS:

- [ ] Classification Plots
- [ ] Chemprop environment investigation nn.Metric.? (remake?)
- [ ] Basic tests

openadmet/models/anvil/workflow.py
```diff
@@ -440,10 +438,10 @@ def run(
         for eval in self.evals:
             # here all the data is passed to the evaluator, but some evaluators may only need a subset
             eval.evaluate(
-                y_true=y_test,
+                y_true=y_test.values,  # Pass as array instead of series
                 y_pred=y_pred,
                 model=self.model,
-                X_train=train_dataloader,
+                X_train=train_dataloader,  # Why are we passing these?
```
hmacdope, line 444: Eval needs the dataloaders if something like hyperparameter sweep and / or CV is being run. 

There may be a better way to do this, open to suggestions.

openadmet/models/architecture/lgbm.py
```diff
@@ -49,3 +50,31 @@ def predict(self, X: np.ndarray) -> np.ndarray:
...
+@models.register("LGBMClassifierModel")
+class LGBMClassifierModel(LGBMModelBase):
+    """
+    LightGBM classification model
+    """
+
+    type: ClassVar[str] = "LGBMClassifierModel"
+    model_class: ClassVar[type] = lgb.LGBMClassifier
+
+    def predict(self, X: np.ndarray) -> np.ndarray:
```
hmacdope, line 74: Shouldn't this be `predict_proba` if we are mirroring sklearn API? I remember you mentioning something about this though could be wrong.

openadmet/models/eval/classification.py
```diff
@@ -0,0 +1,313 @@
...
+                )
+            wandb.log({"metrics": table})
+
+            for metric in self.metric_names:
+                wandb.log({metric: self.data[metric]["value"]})
+
+        self._evaluated = True
+        return self.data
+
+    def stat_and_bootstrap(
```
hmacdope, line 129: Lets move this to a common helper function between regression and classification? As far as I can see its the same right?

openadmet/models/eval/classification.py
```diff
@@ -0,0 +1,313 @@
...
+        if self.use_wandb:
+            artifact = wandb.Artifact(name="metrics_json", type="metric_json")
+            # Add a file to the artifact
+            artifact.add_file(json_path)
+            # Log the artifact
+            wandb.log_artifact(artifact)
+
+
+@evaluators.register("ClassificationPlots")
+class ClassificationPlots(EvalBase):
```
hmacdope, line 201: Very nice

hmacdope (changes requested):
Quick walk through, looking good!

openadmet/models/anvil/workflow.py
smcolby: Gotcha. No ideas currently, but if I end up working on code that utilize these, I'll keep in mind.

openadmet/models/architecture/lgbm.py
smcolby: I thought we'd keep a unified `predict` among implementations. You're right, though, in that `predict_proba` is the useful method for classification (in fact, transforming to match the output of `predict` is pretty trivial: just an argmax over class probabilities). Open to discussing further.

openadmet/models/eval/classification.py
smcolby: Yeah, where should we put it?

openadmet/models/eval/classification.py
smcolby: Thanks!

hmacdope (changes requested):

openadmet/models/eval/classification.py
smcolby: Moved!

openadmet/models/anvil/workflow.py
```diff
@@ -310,7 +309,13 @@ def run(
         logger.info("Model saved")
 
         logger.info("Predicting")
-        y_pred = self.model.predict(X_test_feat)
+        # Check if the model has predict_proba method (classification)
+        if hasattr(self.model, "predict_proba"):
```
hmacdope, line 313: Works but another option is to use a class attribute of the model to mark this e.g `model.is_classifier` or something like that. 

`hasattr` is ok, just would be cleaner with an API touchpoint IMO as only sklearn compatible will have `predict_proba`

openadmet/models/architecture/lgbm.py
```diff
@@ -49,3 +50,31 @@ def predict(self, X: np.ndarray) -> np.ndarray:
         if not self.estimator:
             raise ValueError("Model not trained")
         return self.estimator.predict(X)
+
+
+@models.register("LGBMRegressorModel")
```
hmacdope, line 55: Nice cleanup

openadmet/models/eval/eval_base.py
```diff
@@ -41,3 +43,39 @@ def report(self):
         Report the evaluation
         """
         pass
+
+    def stat_and_bootstrap(
```
hmacdope, line 47: Fantastic

openadmet/models/tests/anvil/test_anvil.py
```diff
@@ -71,3 +76,15 @@ def test_anvil_cross_val_run(tmp_path):
     anvil_spec = AnvilSpecification.from_recipe(basic_anvil_yaml_cv)
     anvil_workflow = anvil_spec.to_workflow()
     anvil_workflow.run(output_dir=tmp_path / "tst")
+
+
+def test_anvil_classification_run(tmp_path):
```
hmacdope, line 81: Very nice

openadmet/models/tests/test_data/basic_anvil_classification.yaml
```diff
@@ -0,0 +1,54 @@
+metadata:
```
hmacdope, line 1: Simply perfect

hmacdope (approved):
A small nitpick, but going to approve and leave at your discretion. Well done, this is a huge upgrade.

Issue #44, opened 2025-02-14
Change base model in `model` to be called `estimator`?


hmacdope:
Currently in our class hierarchy we have a model object with a property `model` that contains the actual estimator like so

```python
class ModelBase(BaseModel, ABC):
    _model: Any = None

    @property
    def model(self):
        return self._model

    @model.setter
    def model(self, value):
        self._model = value
```

This is confusing and ambiguous in many ways and leads to the antipattern of calling `model.model`. 

Perhaps we should re-label this to be `estimator` in the scikit learn fashion.

hmacdope:
Assigning @khuddzu here, will give nice repo tour

PR #103, opened 2025-03-26
Rename model to estimator

khuddzu:
Created to satisfy Issue #44 

I simply changed 

```python
class ModelBase(BaseModel, ABC):
    _model: Any = None

    @property
    def model(self):
        return self._model

    @model.setter
    def model(self, value):
        self._model = value
```

to 

```python
class ModelBase(BaseModel, ABC):
    _model: Any = None

    @property
    def model(self):
        return self._model

    @model.setter
    def estimator(self, value):
        self._model = value
``` 

Let me know if this does not completely satisfy the issue. 

I explored the repo and tried to find at any point where this functionality was called, and could not find any location. Also, the test code continued to run well. Let me know if I have missed anything.

### DIFF
--- a/openadmet/models/architecture/model_base.py
+++ b/openadmet/models/architecture/model_base.py
@@ -28,7 +28,7 @@ class ModelBase(BaseModel, ABC):
         return self._model
 
     @model.setter
-    def model(self, value):
+    def estimator(self, value):
         self._model = value
 
     @abstractmethod
--- a/openadmet/models/architecture/model_base.py
+++ b/openadmet/models/architecture/model_base.py
@@ -29,7 +29,7 @@ class ModelBase(BaseModel, ABC):
 
     @model.setter
     def estimator(self, value):
-        self._model = value
+        self._estimator = value
 
     @abstractmethod
     def from_params(cls, class_params: dict, model_params: dict):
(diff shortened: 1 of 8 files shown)


openadmet/models/architecture/model_base.py
```diff
@@ -28,7 +28,7 @@ def model(self):
         return self._model
 
     @model.setter
-    def model(self, value):
+    def estimator(self, value):
         self._model = value
```
hmacdope, line 32: Change the private attribute also ie to `self._estimator`

khuddzu:
Okay I have now officially changed the property and setter to estimator. 

The following files were edited for new proper use: 
architecture/chemprop.py
architecture/lgbm.py
trainer/sklearn.py
eval/cross_validation.py

I went through and tested that each model, trainer, and eval case ran successfully. However, I am still learning the anvil package so I am not certain of ALL of its capabilities. Let me know if I missed anything. 

PS: I also just noticed that I left a line of code changed in the testing data basic_anvil.yaml file. And I accidentally included this in the commit. I will remove this in a final commit, if all testing passes.

hmacdope (approved):
@khuddzu confirmed this works, merging, congrats on first PR.

Issue #125, opened 2025-04-09
MTENN `models` model implementation


hmacdope:
@khuddzu to implement model class for MTENN

hmacdope:
Done

PR #128, opened 2025-04-10
Mtenn model implementation #125

khuddzu:
Created mtenn_pipeline.py in architecture directory. 

This currently contains untested code. It has a lightning wrapper to define the training steps necessary for schnet specifically, I am not sure if this will work with other MTENN models. And then an MTENNSchnNetModel class, heavily based off of chemprop, but missing some attributes, that I was not certain if necessary. It is worth noting I am trying to use TorchModelBase as well, where chemprop uses PickleModelBase. 

Next step is to create an anvil yaml file to test this method. I assume the data featurizer needs to be complete before I do this? So @hmacdope let me know when that is complete.


openadmet/models/anvil/data_spec.py
```diff
@@ -55,9 +55,9 @@ def read(self) -> tuple[pd.Series, pd.Series]:
 
         # now read the target columns and smiles column
         target = data[self.target_col]
-        smiles = data[self.smiles_col]
+        input = data[self.input_col]
```
hmacdope, line 58: This is the breaking change causing a bunch of tests to fail. I will push some fixes.

openadmet/models/architecture/mtenn.py
```diff
@@ -0,0 +1,86 @@
...
+            for k, v in data.items():
+                data[k] = v.to(self.device)
+            pred,_ = self.model(data)
+            loss = self.loss_fn(pred, target.unsqueeze(0).to(self.device))
+            batch_loss += loss
+
+        avg_loss = batch_loss / len(data_batch)
+        self.log('train_loss', avg_loss)
+        return avg_loss
+
```
hmacdope, line 34: Needs to implement `forward` for `predict` to work, but amazing job on the `PL` wrapper, makes training way easier.

openadmet/models/registries.py
```diff
@@ -13,10 +13,12 @@
 # featurizers
 from openadmet.models.features.molfeat_fingerprint import *  # noqa: F401 F403
 from openadmet.models.features.molfeat_properties import *  # noqa: F401 F403
-from openadmet.models.architecture.chemprop import *  # noqa: F401 F403
+from openadmet.models.features.mtenn import *  # noqa: F401 F403
```
hmacdope, line 16: Just a note here for future reference, you have to add featurizers and models to be imported here for it to register properly for Anvil

openadmet/models/architecture/mtenn.py
```diff
@@ -0,0 +1,86 @@
...
+        Prepare the model
+        """
+        if not self.estimator:
+           #i think scaler is chemprop specific, but if not i will add loop back in
+            model_config = SchNetModelConfig(**self.model_params)
+            self.estimator = MTENNLightningWrapper(model_config)
+        else:
+            logger.warning("Model already exists, skipping build.")
+
+    def from_params(self, params):
```
hmacdope, line 63: Should be easy to implement this, I am also thinking this method might be superfluous in the end, will raise issue.

hmacdope (changes requested):
I pushed a few changes and have a few comments, fantastic work on wrapping with lightning, will make our life a lot easier.

 I have pushed a corresponding set in skunkworks that will allow you to test with `anvil`. eg. `openadmet anvil --recipe-path mtenn_anvil.yaml  --output-dir model0`

khuddzu:
I run into the following error when the model file in architecture is named mtenn.py (was mtenn_pipeline for this reason). 

```
Traceback (most recent call last):
  File "/Users/katehuddleston/Documents/Work/openadmet/openadmet-models/openadmet/models/architecture/mtenn.py", line 5, in <module>
    from mtenn.config import SchNetModelConfig
  File "/Users/katehuddleston/Documents/Work/openadmet/openadmet-models/openadmet/models/architecture/mtenn.py", line 5, in <module>
    from mtenn.config import SchNetModelConfig
ModuleNotFoundError: No module named 'mtenn.config'; 'mtenn' is not a package
```

Are you not running into this issue?

khuddzu:
CI Test failures are just from mtenn dependency mismatch of dgl, which we will remove

devtools/conda-envs/openadmet_models_gpu.yaml
```diff
@@ -39,3 +50,4 @@ dependencies:
   - pip:
     - phx-class-registry
     - email-validator
+    - git+https://github.com/choderalab/mtenn.git@main
```
hmacdope, line 53: Change `@` portion to the branch  with `dgl`

openadmet/models/architecture/mtenn.py
hmacdope: We can just leave this blank for now @khuddzu, see #129. Abstract base class will require it to be implemented but just leave it as `pass`

hmacdope (changes requested):

devtools/conda-envs/openadmet_models_gpu.yaml
khuddzu: i changed this to the correct branch within our openadmet fork

openadmet/models/anvil/data_spec.py
khuddzu: i pushed a fix in test_data that is simply changing smiles_col to input_col. uncertain if the change needs to be more complex, but it appears to be passing now. potentially need more complex tests for the new features

openadmet/models/architecture/mtenn.py
khuddzu: a function named predict_step is actually what was being looked for. I am not sure if forward is being used as is should. i will see what happens if I remove it

khuddzu:
Good news! Anvil works running on cpu now for the 3d mtenn schnet model!! however, I had to make some dirty fixes for this and it really needs cleaning. Next step could be for you to check if it is running well on your end with gpu? 

Note I think the only change you will need to make is this bit in architecture/mtenn.py

```
 with torch.inference_mode():
            acc = "cpu"
            #acc = "auto"       #uncomment line and commet above for gpu usage; if auto doesnt work change string to gpu
            trainer = pl.Trainer(
                logger=None, enable_progress_bar=False, accelerator=acc, devices=1    #changed cpu from auto this can be made back in future workarounds
            )
            preds = trainer.predict(self.estimator, dataloader)
        return torch.cat(preds).numpy().ravel()
```

notice the acc variable. very dirty fix. but it works currently

Also, within mtenn, there are a lot of repetitive steps/loops that can be cleaned up. I also notice that in the end predicting was looking for something called predict_step. I now am realizing I am not sure if forward is being used or not. I will play with removing it and see. I am not sure it needs both.

openadmet/models/anvil/data_spec.py
hmacdope: That should be right!

hmacdope:
@khuddzu somehow the PDB file I added has been lost in a merge somewhere.

You can pull it in from the other PR #123.

Does `auto` work for you out of the box? We can just have that be the default.

hmacdope:
Other than that just fix tests and we are good to go.

khuddzu:
"auto" specifies that you want to use a GPU if a GPU is available, otherwise use a CPU. I am in an interesting situation with my current computer because, as I understand it, the M4 graphics chip is available as a compute GPU. However, it is not really, At least it is not currently Pytorch compatible. So, when I use auto it tries to use the GPU and some things are still sent to my CPU and it gets all messy. I think this is the issue, but I could be wrong.

khuddzu:
Still successful on my end when I remove the forward function. It was looking for predict_step. I did not delete forward for now, just commented it out, in case we run into something else.

khuddzu:
Ready to merge. The only reason I have waited is that the CI/macOS is not finishing for some reason. 

Cleaned up slightly, but still some dirty parts. 
I will create an issue for generally cleaning the mtenn script. An issue has already been created for changes needed for cpu vs gpu computation. Unfortunately "auto" does not work for my computer, therefore I am going to assume this is not a good option for us currently. But the changes needed to make this  work are not a simple clean, therefore i think another PR is needed.

hmacdope (approved):
LGTM @khuddzu congrats and well done.

Issue #106, opened 2025-03-27
Add configurable option for CPU training in lightning


hmacdope:
Related to #105 we should have an option for devices used for Pytorch lightning training.

hmacdope:
@khuddzu re our discussion

PR #133, opened 2025-04-18
Setup for Flexible device selection (gpu vs cpu)

khuddzu:
Satisfies Issue #106 
Making changes to workflow so that user can select gpu or cpu for deep learning pytorch workflows. GPU is set to default, but simply adding cpu to training params in yaml file will allow for cpu usage rather than gpu. 

Done:
- Add accelerator and devices to predict workflow in AnvilDeeplearningWorkflow class. This specifies that we will use the same accelerator and device as we used for the training step. 
- Update MTENNSchNetModel predict function to take accelerator and devices as input variables. 

ToDo:
- Update/check chemprop predict function to align with new workflow capabilities
- Add basic test yaml file for cpu usage

### DIFF
--- a/openadmet/models/anvil/workflow.py
+++ b/openadmet/models/anvil/workflow.py
@@ -436,9 +436,7 @@ class AnvilDeepLearningWorkflow(AnvilWorkflowBase):
         logger.info("Model saved")
 
         logger.info("Predicting")
-        accelerator = getattr(self.trainer, "accelerator", "gpu")
-        devices = getattr(self.trainer, "devices", 1)
-        y_pred = self.model.predict(test_dataloader, accelerator=accelerator, devices=devices)
+        y_pred = self.model.predict(test_dataloader, accelerator=self.trainer.accelerator, devices=self.trainer.devices)
         logger.info("Predictions made")
 
         logger.info("Evaluating")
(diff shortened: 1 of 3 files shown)


openadmet/models/anvil/workflow.py
```diff
@@ -436,7 +436,9 @@ def run(
         logger.info("Model saved")
 
         logger.info("Predicting")
-        y_pred = self.model.predict(test_dataloader)
+        accelerator = getattr(self.trainer, "accelerator", "gpu")
```
hmacdope, line 439: Can you not just use `self.trainer.accelerator.gpu` and `self.trainer.devices`, why use `getattr`?

hmacdope (changes requested):

openadmet/models/anvil/workflow.py
khuddzu: self.trainer.accelerator/self.trainer.devices works. (.gpu shouldnt be there). i will change to this as it is cleaner. 
my thought process on getattr was establishing a default again, especially if someone is only predicting, not training, but that is not necessary and is actually a flawed design now that i think about it.

khuddzu:
Ready to merge. 
@hmacdope  maybe a quick last review? 

I did not add a test for this, as none of the current yaml files in test_data use the LightningModule. I'll add to issues and we can add something that goes through the deep learning tests later.

hmacdope:
Just restarted CI, lets see if it finishes

hmacdope:
Im just gonna merge and we can bugfix later.

PR #136, opened 2025-04-21
Inference CLI 

hmacdope:
Fixes #61 and adds a CLI for inference on models trained with `anvil`

Issue #130, opened 2025-04-11
Use of `to_numpy()` in featurizers is clumsy, should always accept array


hmacdope:
The use of `to_numpy()` in: 

openadmet/models/features/mtenn.py:        y = y.to_numpy().reshape(-1, 1)
openadmet/models/features/chemprop.py:        y = y.to_numpy().reshape(-1, 1)

Is clumsy, as it assumes that the `y` array is a `pandas.Series`. Instead an array should always be passed to the featurizer by the `data` module. 

This should be done carefully as other things may break unexpectedly.

PR #138, opened 2025-04-22
[WIP] Multitask chemprop

### LINKED ISSUE #97: Implement multi-task training for Chemprop
Generally speaking over the ASAP challenge and in several bits of [literature](https://chemrxiv.org/engage/chemrxiv/article-details/672a91bd7be152b1d01a926b) multi-task Chemprop models seemed to outperform their single-task equivalents. This also seems to be the case across many graph-based architectures. 

This is largely because they can share a representation, which improves generalisation. 

Chemprop has native support for multi-tasking, we should use this as a first port of call.

hmacdope:
Fixes #97 

Currently WIP and branching off #136.


openadmet/models/features/chemprop.py
```diff
@@ -23,18 +23,26 @@ def _prepare(self):
...
         Featurize a list of SMILES strings
         """
-        y = y.to_numpy().reshape(-1, 1)
-        dataset = MoleculeDataset(
-            [MoleculeDatapoint.from_smi(smi, y_) for smi, y_ in zip(smiles, y)]
-        )
-        if self.normalize_targets:
-            scaler = dataset.normalize_targets()
+        if y is not None:
+            #fix broken for multi-task
```
hmacdope, line 31: @khuddzu this is currently where I was at, the reshaping is not working for multitask, but need to preserve the behaviour this was originally needed for. 

#130  is also relevant here, we may want to investigate pushing the `to_numpy` call higher up the chain.

khuddzu: Unless I am missing something, I think the fix is as simple as adding a dimensionality check statement. i will add to new commit and we can revert back if it is not what we need

openadmet/models/architecture/chemprop.py
```diff
@@ -114,5 +119,10 @@ def predict(self, X: np.ndarray, accelerator="gpu", devices=1) -> np.ndarray:
                 logger=None, enable_progress_bar=False, accelerator=accelerator, devices=devices
             )
             preds = trainer.predict(self.estimator, X)
-        # concatenate the predictions which are in a list of tensors
-        return torch.cat(preds).numpy().ravel()
+
+       # concatenate the predictions which are in a list of tensors
+        preds = torch.cat(preds, dim=0)
```
hmacdope, line 124: Possibly return 2D by default across all predict?

openadmet/models/tests/test_data/cyp3a4_anvil_lgbm_model_dir/anvil_recipe.yaml
```diff
@@ -0,0 +1,80 @@
+data:
+  anvil_dir: file:///store/projects/OpenADMET_WORK/initial_runs/CYP3A4_cross_val
+  cat_entry: null
+  input_col: OPENADMET_SMILES
+  resource: cyp3a4_chembl_permissive.csv
+  target_col: pchembl_value_mean
```
hmacdope, line 6: Pass target column through to eval, such that labelling is possible and convenient.

hmacdope:
Eval will also need to be made generic, such that it works over multiple target columns, addressing #139 will be critical to this.

openadmet/models/anvil/data_spec.py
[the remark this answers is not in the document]
hmacdope: We could do `Union[str, list[str]]` which would allow both.

hmacdope:
Merged with main here so diff should be much nicer @khuddzu

openadmet/models/architecture/chemprop.py
khuddzu: set 2d default for chemprop and mtenn

khuddzu:
Starting on eval now. Initially it is looking like more restructuring will need to be done than I originally thought. So give me some time on this. I need to play with it and really determine how it works before I make any decisions.

openadmet/models/anvil/data_spec.py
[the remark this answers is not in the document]
khuddzu: yeah thats a good idea

khuddzu:
@hmacdope Could you review for me? Especially the latest commit. I made some changes. 
This allows for the data computed in RegressionMetrics to be fed into RegressionPlots, so that it does not need to be recomputed for plotting. This is optional though. 

Currently only eval_base and regression have been modified, as well as AnvilDeepLearningWorkflow. 
ToDo: Make sure eval is still working with other Anvil workflow class and make changes where necessary in those models.

khuddzu:
Pushed because it is end of day, but still working on getting all tests to pass

khuddzu:
There is one final test that is not passing. I am having trouble finding where it is actually originating from. I am thinking something with shape in classification code, but it is hard if it is something im overlooking, or if it originates from data.

Issue #78, opened 2025-03-13
Add 3 split mode (train, val, test) for deep learning


hmacdope:
We need to be able to operate in a 3 split mode for our deep learning workflows, as 5x5 CV will not always be practical for certain models and datasets. 

A traditional 3 split model will allow us to optimise hyperparameters while not taking a million years.

PR #151, opened 2025-05-13
Enable splitting into train, validation, and test

smcolby:
## Description
For deep learning models, optimal training requires checkpointing against validation set performance. This PR will allow for three-way splits into train, validation, and test sets. 

While not strictly backwards compatible with original two-way splits, requesting train + validation only, or train + test only, is valid usage -- "empty" splits will simply be returned as `None`, and will be handled appropriately by the relevant trainer. Empty validation _and_ test splits will raise a value error (i.e. both cannot be zero).

## Todos
  - [X] Implement three-way split logic
  - [ ] Add additional checkpoint logic
  - [ ] Update unit tests
  - [ ] Ensure functional end-to-end training run (e.g. multitask mtenn)

## Status
- [ ] Ready to go


smcolby:
Resolves #78

openadmet/models/anvil/workflow.py
```diff
@@ -217,39 +302,66 @@ def run(self, output_dir: PathLike = "anvil_run", debug: bool = False) -> Any: .
 
 
 class AnvilWorkflow(AnvilWorkflowBase):
+    """
+    Workflow for running basic Anvil configuration.
+    """
+
     driver: Drivers = Drivers.SKLEARN
 
+    @model_validator(mode="after")
```
hmacdope, line 311: Nice

openadmet/models/anvil/workflow.py
```diff
@@ -453,8 +633,12 @@ def run(
                 y_train=train_dataloader,
                 use_wandb=use_wandb,
                 tag=model_tag,
+                target_labels=target_labels,
             )
+
+            # Write evaluation report
             eval.report(write=True, output_dir=output_dir)
+
         logger.info("Evaluation done")
```
hmacdope, line 643: These all look really good, thanks for commenting everything also

openadmet/models/eval/classification.py
```diff
@@ -74,7 +74,19 @@ def evaluate(self, y_true=None, y_pred=None, use_wandb=False, tag=None, **kwargs
                     _y_pred = y_pred[:, 1].ravel()
                     _y_true = y_true.ravel()
 
-            # Multiclass
+            # Also binary case
+            elif y_true.ndim == 2 and y_true.shape[1] == 1:
```
hmacdope, line 78: Can this not be folded into the branch above? I know its a bit crowded but DRY would say combined them. 

I could be misinterpreting.

openadmet/models/eval/classification.py
```diff
@@ -204,6 +216,10 @@ def roc_curve(
         if y_true.ndim == 1:
             fpr, tpr, _ = roc_curve(y_true.ravel(), y_pred[:, 1].ravel())
 
+        # Also binary case
```
hmacdope, line 219: Same here, can we fold into above branch?

openadmet/models/eval/classification.py
```diff
@@ -234,6 +250,12 @@ def pr_curve(
                 y_true.ravel(), y_pred[:, 1].ravel()
             )
 
+        # Also binary case
```
hmacdope, line 253: Same here

openadmet/models/eval/regression.py
```diff
@@ -33,53 +34,68 @@ class RegressionMetrics(EvalBase):
...
-    def evaluate(self, y_true=None, y_pred=None, use_wandb=False, tag=None, **kwargs):
+    def evaluate(self, y_true=None, y_pred=None, use_wandb=False, tag=None, target_labels=None, **kwargs):
         """
         Evaluate the regression model
         """
         if y_true is None or y_pred is None:
             raise ValueError("Must provide y_true and y_pred")
 
+        n_tasks = y_true.shape[1]
+        assert n_tasks == y_pred.shape[1]
```
hmacdope, line 45: Raise a more descriptive error here.

openadmet/models/eval/regression.py
```diff
@@ -123,13 +146,20 @@ def make_stat_caption(self):
...
-        for metric in self.metric_names:
-            value = self.data[metric]["value"]
-            lower_ci = self.data[metric]["lower_ci"]
-            upper_ci = self.data[metric]["upper_ci"]
-            confidence_level = self.data[metric]["confidence_level"]
-            stat_caption += f"{self._metrics[metric][2]}: {value:.2f}$_{{{lower_ci:.2f}}}^{{{upper_ci:.2f}}}$\n"
-        stat_caption += f"Confidence level: {confidence_level}"
+
+        for task_name in self.task_names:
+            if task_name == 'tag':
```
hmacdope, line 151: What is the role of this?

openadmet/models/eval/regression.py
```diff
@@ -247,6 +294,22 @@ def regplot(
 
         return fig
 
+    def make_stat_caption_from_data(self, data):
```
hmacdope, line 297: This is not used AFAIK? Could be wrong, should be simple to fold in.

openadmet/models/features/chemprop.py
```diff
@@ -27,8 +27,9 @@ def featurize(self, smiles: Iterable[str], y: Iterable[Any]=None) -> DataLoader:
         """
         Featurize a list of SMILES strings
         """
-        if y:
-            y = y.to_numpy().reshape(-1, 1)
+        if y is not None:
+            #fix broken for multi-task
```
hmacdope, line 31: No longer true?

openadmet/models/split/scaffold.py
```diff
@@ -9,45 +9,179 @@ class ScaffoldSplitter(SplitterBase):
...
+            return (
+                X[train_idx],
+                X[val_test_idx],
+                None,
+                y[train_idx],
+                y[val_test_idx],
+                None,
+            )
+
+        # If both test and validation sets are requested, split the remaining data
```
hmacdope, line 48: Nitpick but would prefer this had a guard for its condition, i.e `if self.test_size and self.val_size`

openadmet/models/split/scaffold.py
```diff
@@ -9,45 +9,179 @@ class ScaffoldSplitter(SplitterBase):
...
```
hmacdope, line 110: Same here, not sure if clustering twice is how this is supposed to work, but I think impact should be minimal

hmacdope: Lets raise as an issue.

openadmet/models/split/scaffold.py
```diff
@@ -9,45 +9,179 @@ class ScaffoldSplitter(SplitterBase):
...
```
hmacdope, line 171: Same here with guard

openadmet/models/split/sklearn.py
```diff
@@ -9,15 +9,35 @@ class ShuffleSplitter(SplitterBase):
...
+
+        # If no validation set is requested, return train and test sets
+        if self.val_size == 0:
+            return X_train, None, X_val_test, y_train, None, y_val_test
+
+        # If no test set is requested, return train and validation sets
+        if self.test_size == 0:
+            return X_train, X_val_test, None, y_train, y_val_test, None
+
+        # If both test and validation sets are requested, split the remaining data
```
hmacdope, line 34: Same here

hmacdope (changes requested):
Looking great, just some small changes. :)

openadmet/models/split/scaffold.py
hmacdope: Does it instead make sense to just split the test ratio i.e split into (test, train) and then test becomes -> (test, val)

openadmet/models/eval/classification.py
hmacdope: I have made this change @smcolby, feel free to revert

openadmet/models/eval/regression.py
hmacdope: I have made this change @smcolby

openadmet/models/features/chemprop.py
hmacdope: Fixed

openadmet/models/split/scaffold.py
hmacdope: Can ignore this if you disagree

openadmet/models/eval/classification.py
smcolby: I had this as two if statements because I thought that checking `y_true.shape[1]` would cause an error if `ndim` = 1 (i.e. index out of range), but looks like this usage is fine after a quick check.

openadmet/models/eval/classification.py
smcolby: Propagating above change here.

openadmet/models/eval/regression.py
smcolby: I'm not sure, didn't write this. Looks like it just skips evaluation if the task key is `tag`.

openadmet/models/eval/regression.py
smcolby: Also did not author. Do you suggest removal until we'd make use of it?

openadmet/models/split/scaffold.py
smcolby: I think we should chat briefly on this, not sure I 100% follow

openadmet/models/split/scaffold.py
smcolby: The sum of `test_size` and `val_size` cannot be zero, which is checked at construction in `SplitterBase`. So this branch of the if tree is implicitly guarded for. I made the if:elif:else logic explicit, but did not add additional checks for `test_size` and `val_size`. Sound good?

hmacdope: Yep

openadmet/models/eval/regression.py
hmacdope: Issue raised #171, lets deal with it later.

openadmet/models/split/scaffold.py
hmacdope: Lets push to an issue and discuss later this week.

PR #208, opened 2025-06-20
Add XGBoost regression and classifier models 

### LINKED ISSUE #207: Anvil output directory defaults to `None` if `output-dir` not specified
We should set a sensible default rather than `None`

hmacdope:
## Description
Adds the ability to use XGBoost models for regression and classification. 

Also fixes #207 

## Status
- [X] Ready to go

### DIFF
--- a/openadmet/models/anvil/workflow.py
+++ b/openadmet/models/anvil/workflow.py
@@ -298,7 +298,7 @@ class AnvilWorkflowBase(BaseModel):
     debug: bool = False
 
     @abstractmethod
-    def run(self, output_dir: PathLike = "anvil_run", debug: bool = False) -> Any: ...
+    def run(self, output_dir: PathLike = "anvil_training", debug: bool = False) -> Any: ...
 
 
 class AnvilWorkflow(AnvilWorkflowBase):
@@ -318,7 +318,7 @@ class AnvilWorkflow(AnvilWorkflowBase):
         return self
 
     def run(
-        self, output_dir: PathLike = "anvil_run", debug: bool = False, tag: str = None
+        self, output_dir: PathLike = "anvil_training", debug: bool = False, tag: str = None
     ) -> Any:
         """
         Run the workflow.
@@ -473,7 +473,7 @@ class AnvilDeepLearningWorkflow(AnvilWorkflowBase):
     driver: Drivers = Drivers.PYTORCH
 
     def run(
-        self, output_dir: PathLike = "anvil_run", debug: bool = False, tag: str = None
+        self, output_dir: PathLike = "anvil_training", debug: bool = False, tag: str = None
     ) -> Any:
         """
         Run the workflow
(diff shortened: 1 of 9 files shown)


hmacdope:
@dwwest could I grab your review here when convenient?

openadmet/models/architecture/xgboost.py
```diff
@@ -0,0 +1,80 @@
...
+from openadmet.models.architecture.model_base import PickleableModelBase, models
+
+
+class XGBoostModelBase(PickleableModelBase):
+    """
+    Base class for XGBoost models
+    """
+
+    type: ClassVar[str]
+    mod_class: ClassVar[
```
dwwest, line 16: I'm also guilty of this, but I think we should probably call this "model" since it's just two more letters and in different contexts "mod" could be "modifier"

dwwest: Goes for the other instances of "mod" here

openadmet/models/architecture/xgboost.py
```diff
@@ -0,0 +1,80 @@
...
+from xgboost import XGBClassifier, XGBRegressor
+import numpy as np
+from loguru import logger
+
+from openadmet.models.architecture.model_base import PickleableModelBase, models
+
+
+class XGBoostModelBase(PickleableModelBase):
+    """
+    Base class for XGBoost models
```
dwwest, line 12: Let's get some more filled-out docstrings in here for our model architectures, and add the list of parameters you can input for mod_params and class_params

openadmet/models/architecture/xgboost.py
```diff
@@ -0,0 +1,80 @@
...
+    @classmethod
+    def from_params(cls, class_params: dict = {}, mod_params: dict = {}):
+        """
+        Create a model from parameters
+        """
+        instance = cls(**class_params, mod_params=mod_params)
+        instance.build()
+        return instance
+
+    def train(self, X: np.ndarray, y: np.ndarray):
```
dwwest, line 30: Here, especially, docstrings, X and y are pretty self-explanatory, but let's specify anyway

openadmet/models/anvil/workflow.py
```diff
@@ -298,7 +298,7 @@ class AnvilWorkflowBase(BaseModel):
     debug: bool = False
 
     @abstractmethod
-    def run(self, output_dir: PathLike = "anvil_run", debug: bool = False) -> Any: ...
+    def run(self, output_dir: PathLike = "anvil_training", debug: bool = False) -> Any: ...
```
dwwest, line 301: I like this, much more explicit!

openadmet/models/architecture/xgboost.py
hmacdope: IIRC we were having issues with pydantic protecting the "model_XXX" namespace which resulted in this. I agree its ugly but may need to stay.

openadmet/models/architecture/xgboost.py
hmacdope: Fixed!

hmacdope:
Ready for re-review @dwwest

dwwest (approved):
Looks good!

Issue #218, opened 2025-06-26
Eval figure improvements


hmacdope:
@cynthiaxu04 pointed out that the eval figure (regression plot) needs improvement 

* Make larger 
* Text is hard to read 
* Add density (#201)
* Add error bars (#200), This is more complex as error will need to be fed in from CSV for experimental error, currently limited in doing model error, but should automate long term.

Issue #200, opened 2025-06-10
Add error bars to `Regplots` in both experimental and predicted dimensions.


hmacdope:
We should have error bars available for both the experimental and predicted values (if predictions done with a model ensemble).

PR #223, opened 2025-07-11
Fix/218/improve eval fig

### LINKED ISSUE #201: Add density to axes on `RegPlots`
We should add a seaborne style density estimator (likely KDE) on the axes of the regression plots.

cynthiaxu04:
## Description
- partially addresses #218 
- makes final regplot figure larger
- adds density to regplot in the form of histograms (kdeplot is not possible because of an error in an older version of seaborn, but tabfpn requires that older version of seaborn, so this is a workaround); Closes #201 
- adds `ciplot` which plots the regression metrics with confidence intervals as error bars; exports as `cross_validation_ciplot.png`
- error bars from experimental error feature still needs to be added #200

## Todos
  - [ ] TODO 1

## Questions
- [ ] Question1

## Status
- [ ] Ready to go

### DIFF
--- a/openadmet/models/eval/cross_validation.py
+++ b/openadmet/models/eval/cross_validation.py
@@ -21,6 +21,8 @@ from openadmet.models.eval.regression import (
     nan_omit_spearmanr,
 )
 from openadmet.models.trainer.lightning import LightningTrainer
+from openadmet.models.eval.utils import _make_stat_caption, _make_stat_dict
+
 
 
 def wrap_ktau(y_true, y_pred):
@@ -173,13 +175,13 @@ class SKLearnRepeatedKFoldCrossValidation(CrossValidationBase):
         self.plot_data = {}
 
         # stat_caption = self.make_stat_caption(t_label)
-        stat_dict = self.make_stat_dict(t_label)
+        stat_dict = self.get_stat_dict(t_label=t_label)
 
         # create the plots
         for plot_tag, plot in self.plots.items():
             if "ciplot" in plot_tag:
                 self.plot_data[plot_tag] = plot(stat_dict=stat_dict)
-            else:
+            elif "regplot" in plot_tag:
                 self.plot_data[plot_tag] = plot(
                     y_true,
                     y_pred,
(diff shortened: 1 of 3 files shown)


openadmet/models/eval/cross_validation.py
```diff
@@ -440,22 +476,27 @@ def evaluate(
...
+                        t_pred,
+                        xlabel=self.axes_labels[0],
+                        ylabel=self.axes_labels[1],
+                        title=f"{self.title}\nTask: {t_label}",
+                        # stat_caption=stat_caption,
+                        stat_dict=stat_dict,
+                        pXC50=self.pXC50,
+                        min_val=self.min_val,
+                        max_val=self.max_val,
+                    )
```
smcolby, line 500: Same here for if:else block

openadmet/models/eval/cross_validation.py
```diff
@@ -507,3 +548,32 @@ def make_stat_caption(self, task_name):
...
+            upper_ci = self.data[task_name][metric]["upper_ci"]
+            confidence_level = self.data[task_name][metric]["confidence_level"]
+
+            # Save in dict
+            stat_dict["metrics"].append(self._metrics[metric][2])
+            stat_dict["means"].append(value)
+            stat_dict["lower_ci"].append(lower_ci)
+            stat_dict["upper_ci"].append(upper_ci)
+            stat_dict["conf_level"] = confidence_level
+        return stat_dict
```
smcolby, line 579: If `make_stat_dict` is identical for the sklearn and lightning implementations, could move to a helper function `_make_stat_dict` and call in both classes. Reduces repeated code.

openadmet/models/eval/regression.py
```diff
@@ -183,6 +183,41 @@ def make_stat_caption(self):
...
+            confidence_level = self.data[task_name][metric]["confidence_level"]
+
+            # Save in dict
+            stat_dict["metrics"].append(self._metrics[metric][2])
+            stat_dict["means"].append(float(value))
+            stat_dict["lower_ci"].append(float(lower_ci))
+            stat_dict["upper_ci"].append(float(upper_ci))
+            stat_dict["conf_level"] = confidence_level
+        return stat_dict
+
```
smcolby, line 218: Another reason to have a `_make_stat_dict` helper such that it can be maintained in only one place.

cynthiaxu04: helper functions added!

openadmet/models/eval/cross_validation.py
cynthiaxu04: helper functions added!

openadmet/models/eval/cross_validation.py
```diff
@@ -167,25 +167,31 @@ def evaluate(
...
-                y_pred,
-                xlabel=self.axes_labels[0],
-                ylabel=self.axes_labels[1],
-                title=f"{self.title}\nTask: {t_label}",
-                stat_caption=stat_caption,
-                pXC50=self.pXC50,
-                min_val=self.min_val,
-                max_val=self.max_val,
-            )
+            if "ciplot" in plot_tag:
```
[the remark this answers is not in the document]
cynthiaxu04: see below comment

hmacdope (changes requested):
Huge improvement over our existing plots. 

Just some nitpicks to work through but then should be good to go

PR #224, opened 2025-07-14
Fix/215/sklearn mask nan

### LINKED ISSUE #215: Traditional ML workflow should mask `NaN` effectively
As @cynthiaxu04 observed, currently the LGBM and similar scikit-learn driven workflows jam when `NaN` values are in the targets. We should make sure this works automatically like with chemprop workflows

cynthiaxu04:
## Description
Closes #215. 

## Todos
  - [x] `DataSpec.read()` updated to drop `NaN`s for both Traditional ML and DL anvil workflows
  - [x] `DataSpec.read()` also now returns the original number of rows and the number of rows dropped if any. This will appear in a logger.info message

## Questions
- [ ] Question1

## Status
- [ ] Ready to go


openadmet/models/anvil/data_spec.py
```diff
@@ -68,7 +68,25 @@ def read(self) -> tuple[pd.Series, pd.Series]:
...
+            n_dropped = n_before - n_after
+        else:
+            n_dropped = 0
+            cleaned_combined = combined
+
+        # Split the data again
+        input_clean = cleaned_combined[self.input_col]
+        targets_clean = cleaned_combined[self.target_cols]
+
+        return input_clean, targets_clean, n_dropped, n_before
```
smcolby, line 89: Why are we returning so much here? My expectation would be to get back X (inputs) and y (targets). Moreover, I think we generally want to broker in numpy arrays when passing data around, not pandas Series.

@hmacdope might have opinions here.

openadmet/models/anvil/data_spec.py
```diff
@@ -68,7 +68,25 @@ def read(self) -> tuple[pd.Series, pd.Series]:
         targets = data[self.target_cols]
         input = data[self.input_col]
 
-        return input, targets
+        # combine input and targets for joint NaN handling
+        combined = pd.concat([input, targets], axis=1)
+
```
smcolby, line 73: Could be replaced by:
```
combined = data[self.input_cols + self.target_cols]
```
You're otherwise subsetting the data and recombining.

openadmet/models/anvil/workflow.py
```diff
@@ -381,7 +381,8 @@ def run(
 
         # Load data from YAML specification
         logger.info("Loading data")
-        X, y = self.data_spec.read()
+        X, y, n_dropped, n_before = self.data_spec.read()
+        logger.info(f"{n_before} total rows. {n_dropped} NaN rows were dropped.")
```
smcolby, line 385: This might answer my question: `n_dropped` and `n_before` are used for logging, but this info is only accessible if it "comes out of" `data_spec.read()`. 

Feels a bit hacky to save this information out just to print a log message, particularly when `dropna` isn't necessarily going to be set to `True`. Might I suggest we add dropping NaNs as a data filter and separate step, rather than trying to force it into the load functionality? Or simply log "Dropping NaN values" without a count in the output?

openadmet/models/anvil/workflow.py
```diff
@@ -381,7 +381,8 @@ def run(
 
         # Load data from YAML specification
         logger.info("Loading data")
-        X, y = self.data_spec.read()
+        X, y, n_dropped, n_before = self.data_spec.read()
```
smcolby, line 384: `data_spec.read()` is not passed a `dropna` argument, so NaNs won't ever be dropped here.

openadmet/models/tests/data/test_data.py
```diff
@@ -23,6 +23,6 @@ def test_data_spec_from_intake():
         target_cols=["data1"],
         input_col="SMILES",
     )
-    target, smiles = data_spec.read()
+    target, smiles, _, _ = data_spec.read()
     assert len(target) == 30
     assert len(smiles) == 30
```
smcolby, line 28: Should have tests for when `dropna` is `True`

smcolby (changes requested):

openadmet/models/anvil/data_spec.py
hmacdope: See below comment on logging

openadmet/models/anvil/workflow.py
cynthiaxu04: `dropna` is defaulted to True

openadmet/models/anvil/data_spec.py
```diff
@@ -64,11 +66,28 @@ def read(self) -> tuple[pd.Series, pd.Series]:
...
         else:
             raise ValueError(f"Unsupported resource type: {self.resource}")
-        # now read the target columns and smiles column
-        targets = data[self.target_cols]
-        input = data[self.input_col]
 
-        return input, targets
+        # combine input and targets for joint NaN handling
+        combined = data[[self.input_col] + (self.target_cols if isinstance(self.target_cols, list) else [self.target_cols])]
+
```
smcolby, line 72: Nice.

openadmet/models/anvil/data_spec.py
```diff
@@ -64,11 +66,28 @@ def read(self) -> tuple[pd.Series, pd.Series]:
...
+        # get number of combined rows for logging
+        n_before = len(combined)
+
+        if self.dropna:
+            cleaned_combined=combined.dropna()
```
smcolby, line 77:
```suggestion
            cleaned_combined = combined.dropna()
```

openadmet/models/anvil/data_spec.py
```diff
@@ -64,11 +66,28 @@ def read(self) -> tuple[pd.Series, pd.Series]:
...
+            n_dropped = 0
+            cleaned_combined = combined
+
+        # Split the data again
+        input_clean = cleaned_combined[self.input_col]
+        targets_clean = cleaned_combined[self.target_cols]
+
+        logger.info(f"{n_before} total rows. {n_dropped} NaN rows were dropped.")
+
+        return input_clean, targets_clean
```
smcolby, line 90: Do we want these returned as data frames / series, or as numpy arrays? I.e. `input_clean.values, targets_clean.values`.

openadmet/models/tests/data/test_data.py
```diff
@@ -26,3 +26,26 @@ def test_data_spec_from_intake():
...
+        input_col="OPENADMET_CANONICAL_SMILES",
+        dropna=False
+    )
+
+    target, smiles = data_spec.read()
+    target2, smiles2, = data_spec2.read()
+    assert len(target) == 3333
+    assert len(smiles) == 3333
+    assert len(target2) == 7196
+    assert len(smiles2) == 7196
```
smcolby, line 51: This will test both of your cases independently with fewer lines of code.

```python
@pytest.mark.parametrize("dropna, expected_length", [(True, 3333), (False, 7196)])
def test_data_spec_dropna(dropna, expected_length):
    data_spec = DataSpec(
        type="intake",
        resource=nan_data,
        target_cols=["OPENADMET_LOGAC50"],
        input_col="OPENADMET_CANONICAL_SMILES",
        dropna=dropna
    )

    target, smiles = data_spec.read()

    assert len(target) == expected_length
    assert len(smiles) == expected_length
```

smcolby (changes requested):
Almost there!

openadmet/models/anvil/data_spec.py
cynthiaxu04: the read function has the return defined as a tuple of pandas series

smcolby: I'm referring to #130 here

smcolby (approved):
All good!

openadmet/models/anvil/data_spec.py
cynthiaxu04: has to do with issue #130

PR #227, opened 2025-07-15
Split into train+val and test, then train, val, and test

### LINKED ISSUE #172: Clustering twice in structure based splitters for test and val split.
Same here, not sure if clustering twice is how this is supposed to work, but I think impact should be minimal

_Originally posted by @hmacdope in https://github.com/OpenADMET/openadmet-models/pull/151#discussion_r2094660331_

### LINKED ISSUE #225: Split into train+val and test, then separate train and val if needed
Currently we split into train and val+test, then separate val and test if needed. Thus, if we compare a model with a validation split and one without, the test splits will not be comparable.

smcolby:
## Description
Previously, splits were performed as train/val+test, then train/val/test. This results in inconsistent training sets when comparing models that use versus don't use a validation split (e.g. ChemProp versus LGBM). This PR ensures test splits are canonical.

Additionally, modifies structured splitters to split on structure for train+val/test, but simple random sampling to further split train and val.

Resolves #225
Resolves #172

## Todos
  - [X] Change train+val/test to train/val+test
  - [X] Update return logic for edge cases (all train, no val, no test)
  - [X] Modify structured splitters to split train/val+test eponymously; shuffle split for train/val/test.

## Status
- [X] Ready to go

### DIFF
--- a/openadmet/models/split/sklearn.py
+++ b/openadmet/models/split/sklearn.py
@@ -14,31 +14,35 @@ class ShuffleSplitter(SplitterBase):
         Split the data
         """
 
-        # First split into train and val+test
-        X_train, X_val_test, y_train, y_val_test = train_test_split(
+        # No test nor validation set requested, return the entire dataset as train
+        if self.test_size == 0 and self.val_size == 0:
+            return X, None, None, y, None, None
+
+        # Split into train+val and test
+        X_train_val, X_test, y_train_val, y_test = train_test_split(
             X,
             y,
             train_size=None,
-            test_size=int((self.test_size + self.val_size) * X.shape[0]),
+            test_size=int(self.test_size * X.shape[0]),
             random_state=self.random_state,
         )
 
-        # If no validation set is requested, return train and test sets
+        # No validation set requested, return train and test sets
         if self.val_size == 0:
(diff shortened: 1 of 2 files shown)


openadmet/models/split/sklearn.py
```diff
@@ -14,31 +14,39 @@ def split(self, X, y):
...
+        if self.test_size == 0:
+            # Split into train and val
+            X_train, X_val, y_train, y_val = train_test_split(
+                X,
+                y,
+                train_size=None,
+                test_size=int(self.val_size * X.shape[0]),
+                random_state=self.random_state,
+            )
+            return X_train, X_val, None, y_train, y_val, None
```
khuddzu, line 27: Not necessary for now, but just a thought. Would it overall be cleaner to define some variable defining the splits desired that can have either size 2 or 3? For example, [0.8, 0.2] vs [0.7, 0.15, 0.15] and the order is always the 0th indices is train, and the subsequent are always in the order of val, test, if the len is 3? 
This could in the end be messier, once implemented. But my initial thought is less variables, and you wouldn't have to have multiple lines of code where we are doing the same thing, with different variable names. 
Let me know if that makes sense to you.

PR #234, opened 2025-07-17
Add nepare model to architectures

dwwest:


openadmet/models/features/pairwise.py
```diff
@@ -0,0 +1,66 @@
+import numpy as np
```
hmacdope, line 1: Change module name to be `nepare.py`, more specific.

openadmet/models/registries.py
```diff
@@ -28,6 +30,8 @@
 from openadmet.models.architecture.xgboost import *  # noqa: F401 F403
 from openadmet.models.architecture.catboost import *  # noqa: F401 F403
 from openadmet.models.architecture.mtenn import *  # noqa: F401 F403
+from openadmet.models.architecture.nepare import *  # noqa: F401 F403
```
hmacdope, line 33: Needed imports here

openadmet/models/features/pairwise.py
```diff
@@ -0,0 +1,68 @@
...
+        y: ArrayLike = None,
+    ) -> tuple[DataLoader, np.ndarray, StandardScaler, Dataset]:
+        """
+        Featurize a list of SMILES strings. Returns a DataLoader, a list of indices that correspond to the original input, a StandardScaler if any scaling done by featurization, and a Pytorch Dataset
+        """
+        featurizer = ChemPropFeaturizer(n_jobs=self.n_jobs, normalize_targets=False, batch_size=self.batch_size, shuffle=self.shuffle)
+        _, _, _, unpaired_dataset = featurizer.featurize(smiles, y)
+
+        X = np.array(unpaired_dataset.smiles)
+        y = np.array([dp.y[0] for dp in unpaired_dataset.data]) # assumes target is a single value
```
hmacdope, line 55: No need for this part, just use `PairwiseAugmentedDataset` directly as you do on L57 without featurizing with `ChemProp` first.

openadmet/models/architecture/nepare.py
```diff
@@ -0,0 +1,41 @@
...
+    def validation_step(self, batch: tuple[torch.Tensor, torch.Tensor], batch_idx):
+        return self._step(batch, "validation")
+
+    def test_step(self, batch: tuple[torch.Tensor, torch.Tensor], batch_idx):
+        return self._step(batch, "testing")
+
+    def predict_step(self, batch: tuple[torch.Tensor, torch.Tensor, torch.Tensor]):
+        x_1, x_2, _ = batch
+        x = torch.cat((x_1, x_2), dim=1)
+        return self(x)
```
hmacdope, line 41: Here there is some confusion between `Model` and `Module`. Model = wrapper for our framework. Module = Lightning implementation of  a model architecture. 

@JacksonBurns has implemented the lightning logic here: https://github.com/JacksonBurns/neural-pairwise-regression/blob/main/nepare/nn.py#L43

We need to wrap that up ourselves. 

So the main API points you need to implement are

- make_new
- train
- **build**
- **predict**

With build and predict being the most important. The core of the implementation is **build** where you should construct **.estimator** using the lightning abstractions we already have e.g 

```python 
from nepare.nn import NeuralPairwiseRegressor # this is already a lightning module, implementing `training_step` etc.

class NepareModel(...):
...

def build(self, scale=None):
     ...
     self.estimator = NeuralPairwiseRegressor(self.input_size, self.hidden_size, self.num_layers, self.activation, self.n_targets)
    ...
```
You also shouldn't bother with inheriting from ChemPropModel for now its too overloaded with stuff already. Just implement each API point for `nepare` subclassing `LightningModelBase` instead.

hmacdope (changes requested):
See comments, also needs a GPU integration test and single epoch CPU test through anvil.

openadmet/models/architecture/nepare.py
```diff
@@ -4,14 +4,63 @@
...
+from typing import OrderedDict
 
 from openadmet.models.architecture.model_base import models as model_registry
 from openadmet.models.architecture.model_base import LightningModelBase
 
-from nepare.nn import NeuralPairwiseRegressor
 
 from typing import ClassVar
 
+class NeuralPairwiseRegressorModule(pl.LightningModule):
```
hmacdope, line 15: Reason that we vendor this?

dwwest: This is at @smcolby's suggestion, because the base nepare module doesn't include everything we need, mainly the monitor_metric, see [here](https://openadmet.slack.com/archives/C08CNFCBG3D/p1756933384803559)

hmacdope (changes requested):
This isn't resilient to changes in `input_size` yet, will need checking of why dimensions are not lining up.

dwwest:
In testing inference, I found an issue where if you try to pass a parquet file, it doesn't error, I added the appropriate check

hmacdope:
@smcolby would you be able to review here also so that we can check compliance with new way of eliminating mod_params in #338

openadmet/models/features/pairwise.py
```diff
@@ -0,0 +1,219 @@
...
+        feat = self.featurizer
+        if hasattr(feat, "__class__"):
+            feat_type = feat.__class__.__name__
+            # Remove private fields and methods from dict
+            feat_params = {
+                k: v for k, v in feat.__dict__.items() if not k.startswith("_")
+            }
+            featurizer_field = {feat_type: feat_params}
+        else:
+            featurizer_field = feat
```
smcolby, line 214: I think `model_dump` does this for you, such that you can just call `self.__class__(**self.model_dump())` below.

openadmet/models/anvil/workflow.py
```diff
@@ -412,7 +414,9 @@ def check_if_val_needed(self):
 
         return self
 
-    def _train(self, train_dataloader, val_dataloader, train_scaler, output_dir):
+    def _train(
+        self, train_dataloader, val_dataloader, train_scaler, output_dir, **kwargs
+    ):
```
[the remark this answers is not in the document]
dwwest: GAT also definitely needs this, that's true. I wasn't sure if we wanted the scaler outside of this, but I will add it in now!

openadmet/models/features/pairwise.py
```diff
@@ -0,0 +1,219 @@
...
+
+    def __getitem__(self, index):
+        """Return the item at the given index."""
+        i, j = self.idxs[index]
+        if self.y is not None:
+            return self.X[i], self.X[j], self.y[i] - self.y[j]
+        else:
+            return self.X[i], self.X[j]
+
+    def downsample_(self, n: int, random_seed: int = 1701):
```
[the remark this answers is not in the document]
dwwest: No idea... I think this is an artifact from original nepare... maybe it's Jackson's favorite number?

openadmet/models/anvil/workflow.py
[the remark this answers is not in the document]
dwwest: Actually, I am going to leave this for now, for the sake of getting v1 out on time, but I'll make a separate issue to reconcile all arguments that should be encapsulated by kwargs

openadmet/models/architecture/nepare.py
```diff
@@ -0,0 +1,326 @@
...
+            Concatenated predictions for the batch.
+
+        """
+        if len(batch) == 3:
+            x_1, x_2, _ = batch
+        else:
+            x_1, x_2 = batch
+        x = torch.cat((x_1, x_2), dim=1)
+        y_hat = self(x)
+        return y_hat
```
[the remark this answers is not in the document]
dwwest: I think because you're not given ys in the predict step.

It's consisten with what Jackson had for nepare:
https://github.com/JacksonBurns/neural-pairwise-regression/blob/main/nepare/nn.py

openadmet/models/eval/regression.py
```diff
@@ -109,10 +114,7 @@ def evaluate(
             self.use_wandb = use_wandb
 
         for task_id in range(n_tasks):
-            t_true = y_true[:, task_id]
-            t_pred = y_pred[:, task_id]
-            # remove Nan values
-            t_true, t_pred = mask_nans(t_true, t_pred)
```
hmacdope, line 115: @dwwest I think the loss of `mask_nans` is responsible for the failing test here. Be careful! I think order matters here IIRC.

openadmet/models/eval/cross_validation.py
```diff
@@ -629,14 +630,11 @@ def evaluate(
 
         # now the plots
         for task_id in range(n_tasks):
-            t_true = y_true[:, task_id]
-            t_pred = y_pred[:, task_id]
-            # remove Nan values
-            t_true, t_pred = mask_nans(t_true, t_pred)
```
hmacdope, line 635: Same here, loss of `mask_nans` is the issue here.

openadmet/models/eval/eval_base.py
```diff
@@ -83,6 +84,63 @@ def mask_nans_std(y_true: np.ndarray, y_pred: np.ndarray, y_std: np.ndarray):
...
+                for i in range(N)
+                for j in range(N)
+            ]
+        )
+        t_pred = y_pred[:, task_id]
+        logger.warning(
+            f"Generated {t_true.shape[0]} pairwise differences for task {task_id}"
+        )
+        # Generate a random sample of 1000 indices
+        sample_indices = np.random.choice(len(t_true), size=5000, replace=False)
```
hmacdope, line 127: I think I'm missing something here, but why is this stochastic? Perhaps I just need it explained.

hmacdope (changes requested):

openadmet/models/eval/eval_base.py
dwwest: So this is the only place where all pairs are generated and saved as an object, and it slows the eval down to the point it's unusable. This seemed like the easiest/best patch for now.

smcolby (changes requested):
Small stuff there and there

openadmet/models/architecture/nepare.py
```diff
@@ -73,6 +73,10 @@ def __init__(
         self.lr = lr
         self.save_hyperparameters()
 
+    def __hash__(self):
+        """Hash based on object id to ensure uniqueness in sets and dicts."""
+        return id(self)
```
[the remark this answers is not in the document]
dwwest: There's something in I think eval that uses the type of the architecture class as a hash for a dictionary, and this is handled in the pydantic cases. This is to handle that

openadmet/models/features/pairwise.py
```diff
@@ -0,0 +1,204 @@
...
+        how_to_pair = values.get("how_to_pair")
+        if how_to_pair not in ["full", "ut", "sut"]:
+            raise ValueError("how_to_pair must be one of 'full', 'ut', or 'sut'")
+        return values
+
+    @field_validator("featurizer", mode="before")
+    @classmethod
+    def validate_featurizer(cls, value):
+        """If passed a dictionary of parameters, construct the relevant featurizer and return it."""
+        if isinstance(value, dict):
```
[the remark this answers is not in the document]
dwwest: Yeah, it was supposed to be for dealing with anvil inputs...

PR #241, opened 2025-07-23
Fixes chemprop y_true indexing error in cv

dwwest:


dwwest:
Confirming, this does run with Chemprop

hmacdope (approved):

PR #250, opened 2025-07-27
unimol added and test

yxnyu:
## Description
just check the envi of unimol


hmacdope:
@yxnyu environment seems to work now

hmacdope (changes requested):
@yxnyu you can update this PR with additional environment file that adds unimol + in-repo unimol binding code when ready.

PR #274, opened 2025-08-07
Add dummy classification and regression models

hmacdope:
## Description
Adds dummy models for classification and regression that predict the mean and most populous class respectively. 

## Status
- [ ] Ready to go

### DIFF
--- a/openadmet/models/architecture/dummy.py
+++ b/openadmet/models/architecture/dummy.py
@@ -82,7 +82,7 @@ class DummyRegressorModel(DummyModelBase):
     Dummy regression model
 
     Common parameters for dummy models can be found at:
-    https://scikit-learn.org/stable/modules/dummy.html
+    https://scikit-learn.org/stable/api/sklearn.dummy.html
     """
     type: ClassVar[str] = "DummyRegressorModel"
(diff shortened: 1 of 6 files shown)


hmacdope:
@khuddzu can I get a review here

openadmet/models/architecture/dummy.py
```diff
@@ -0,0 +1,107 @@
...
+
+
+
+@models.register("DummyClassifierModel")
+class DummyClassifierModel(DummyModelBase):
+    """
+    Dummy classification model
+
+    Common parameters for dummy models can be found at:
+    https://scikit-learn.org/stable/modules/dummy.html
```
khuddzu, line 103: same with this one

khuddzu:
Do you have a test use case for this that you can share? There are a couple of lines that based on looking I am not sure how they play out, and would like to play with running it before I give the ole check of approval. Off the bat it looks good, but im wondering if there are not redundancies, or overwriting.

hmacdope:
Block, tests needed

hmacdope:
@khuddzu should be ready for re-review

hmacdope:
Im just going to merge here.

Issue #284, opened 2025-08-13
[BUG] Use of bleeding edge TabPFN is failing due to AutoGluon changes


hmacdope:
Our bleeding edge install of TabPFN and TabPFN-extensions is not working as of tabpfn-extensions 0.1.3 in which they move to an AutoGluon based setup. 

`NaN` values seem to be being inserted into candidate dataframes.

PR #294, opened 2025-08-18
Switch to default TabPFN rather than buggy TabPFN-extensions interface

hmacdope:
## Description
Partial fix for #284, moving to the default TabPFN shipped with  the main repo, not the `tabpfn-extensions` interface. 

## Status
- [X] Ready to go

### DIFF
--- a/openadmet/models/architecture/tabpfn.py
+++ b/openadmet/models/architecture/tabpfn.py
@@ -56,7 +56,7 @@ class TabPFNExtensionModelBase(PickleableModelBase):
         """
         if value not in ["cpu", "gpu", "auto"]:
             raise ValueError("Accelerator must be either 'cpu' or 'gpu' or 'auto'")
-        
+
         return value
 
 


openadmet/models/architecture/tabpfn.py
```diff
@@ -124,3 +125,68 @@ def predict_proba(self, X: np.ndarray) -> np.ndarray:
         if not self.estimator:
             raise ValueError("Model not trained")
         return self.estimator.predict_proba(X)
+
+
+
+class TabPFNModelBase(PickleableModelBase):
+    device: Literal["cpu", "cuda", "auto"] = Field(default="auto")
```
smcolby, line 132: This is breaking with our use of `accelerator` and "cpu" versus "gpu". Do we want to enforce consistency?

openadmet/models/architecture/tabpfn.py
```diff
@@ -124,3 +125,68 @@ def predict_proba(self, X: np.ndarray) -> np.ndarray:
...
+    def from_params(cls, class_params: dict = {}, mod_params: dict = {}):
+        """
+        Create a model from parameters
+        """
+
+        instance = cls(**class_params, mod_params=mod_params)
+        instance.build()
+        return instance
+
+
```
smcolby, line 146:
```suggestion
```

openadmet/models/architecture/tabpfn.py
```diff
@@ -124,3 +125,68 @@ def predict_proba(self, X: np.ndarray) -> np.ndarray:
...
+class TabPFNRegressorModel(TabPFNModelBase):
+    """
+    TabPFN regression model using the basic `tabpfn` implementation.
+    """
+
+    type: ClassVar[str] = "TabPFNRegressorModel"
+    mod_class: ClassVar[type] = TabPFNRegressor
+
+
+
```
smcolby, line 184:
```suggestion
```

openadmet/models/architecture/tabpfn.py
```diff
@@ -124,3 +125,68 @@ def predict_proba(self, X: np.ndarray) -> np.ndarray:
...
+class TabPFNModelBase(PickleableModelBase):
+    device: Literal["cpu", "cuda", "auto"] = Field(default="auto")
+    random_state: int = Field(default=42)
+    ignore_pretraining_limits: bool = Field(default=False)
+
+    @classmethod
+    def from_params(cls, class_params: dict = {}, mod_params: dict = {}):
```
smcolby, line 137: I know you're trying to get away from `from_params`. Should we instead add a custom `__init__` method?

```python
def __init__(self, **kwargs):
    # Call the pydantic-generated init method
    super().__init__(**kwargs)

    # Call `build` to fully initialize instance
    self.build()
```

openadmet/models/architecture/tabpfn.py
```diff
@@ -9,7 +10,7 @@
 
 
 
-class TabPFNModelBase(PickleableModelBase):
+class TabPFNExtensionModelBase(PickleableModelBase):
```
smcolby, line 13: PR says we're moving away from TabPFN extensions, yet here we have `TabPFNExtensionModelBase`. I'm not up on what the differences are, but want to make sure "extension" here isn't an oversight.

smcolby (changes requested):
Some minor things to resolve and/or decide upon.

openadmet/models/architecture/tabpfn.py
hmacdope: This is how TabPFN does it but I think switching to `accelerator` is right call.

openadmet/models/architecture/tabpfn.py
hmacdope: Sorry, wasn't clear in title

Two options now available once this PR merged

1. Use normal TabPFN
2. Use TabPFN extensions interface which does "posthoc ensembling" automatically

smcolby: Got it!

openadmet/models/architecture/tabpfn.py
```diff
@@ -46,14 +47,14 @@ class TabPFNModelBase(PickleableModelBase):
...
-    def validate_device(cls, value):
+    def validate_accelerator(cls, value):
         """
-        Validate the device parameter
+        Validate the accelerator parameter
         """
         if value not in ["cpu", "cuda", "auto"]:
-            raise ValueError("Device must be either 'cpu' or 'cuda' or 'auto'")
+            raise ValueError("Accelerator must be either 'cpu' or 'cuda' or 'auto'")
         return value
```
smcolby, line 58: Should be "gpu", not "cuda"

openadmet/models/architecture/tabpfn.py
smcolby: Roger that.

smcolby (changes requested):
Just needs "gpu" instead of "cuda"!

Issue #325, opened 2025-09-08
Expose uncertainty calibration method in `EnsembleSpec`


smcolby:
There is currently no way to select between `scaling-factor` and `isotonic-regression` for uncertainty calibration in workflows.

PR #328, opened 2025-09-10
Add uncertainty calibration to evaluation pipeline

smcolby:
## Description
If training an ensemble, we should be able to output calibration plots (before and after) as well as the appropriate metrics as an `Eval`

## Todos
  - [x] Expose uncertainty calibration method in `EnsembleSpec`
  - [x] Metrics
  - [x] Plots
  - [x] Tests
  - [x] Docstrings

## Status
- [x] Ready to go


smcolby:
Should resolve #325 here while we're at it

khuddzu (approved):

PR #336, opened 2025-09-17
Implement sphinx autodocs for API doc

dwwest:


hmacdope (approved):
@dwwest I changed the API doc layout so that the toctrees work. Try building it yourself and see if you like how it works. It's a bit more fiddly but I think neater. We should follow a similar pattern in rest of docs. 

When you are happy, go ahead and merge.

PR #338, opened 2025-09-17
Enable freezing select model weights

### LINKED ISSUE #337: Enable weight freezing for portions of network architectures
Fine-tuning pipeline currently is able to load network weights and continue training, but all parameters remain fully trainable. We should be able to designate some or all network components as "frozen", such that weights are no longer updated for these portions.

smcolby:
## Description
Fine-tuning pipeline currently is able to load network weights and continue training, but all parameters remain fully trainable. We should be able to designate some or all network components as "frozen", such that weights are no longer updated for these portions.

Will resolve #337 

## Todos
  - [x] Implement ability to freeze weights, per architecture
  - [x] Expose relevant options in anvil workflow
  - [x] Tests

## Status
- [x] Ready to go

### DIFF
--- a/openadmet/models/architecture/model_base.py
+++ b/openadmet/models/architecture/model_base.py
@@ -334,16 +334,6 @@ class LightningModuleBase(pl.LightningModule):
         else:
             return optimizer
 
-    # TODO: Implement defaults of the following?
-    # def training_step():
-    #     pass
-
-    # def validation_step():
-    #     pass
-
-    # def predict_step():
-    #     pass
-
 
 class LightningModelBase(ModelBase):
     """A model that uses PyTorch Lightning."""
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -37,7 +37,7 @@ repos:
     - --py39-plus
 - repo: https://github.com/astral-sh/ruff-pre-commit
   # Ruff version.
-  rev: v0.13.1
+  rev: v0.13.2
   hooks:
     # Run the linter.
     - id: ruff-check
--- a/openadmet/models/active_learning/committee.py
+++ b/openadmet/models/active_learning/committee.py
@@ -81,10 +81,6 @@ class CommitteeRegressor(EnsembleBase):
         **kwargs : dict
             Additional keyword arguments to be passed to the committee's predict method.
 
-        Returns
-        -------
-        None
-
         """
         # Reset calibration model
         self._calibration_model = None
@@ -128,10 +124,6 @@ class CommitteeRegressor(EnsembleBase):
         **kwargs : dict
             Additional keyword arguments to be passed to the committee's predict method.
 
-        Returns
-        -------
-        None
-
         """
         # Reset calibration model
         self._calibration_model = None
@@ -168,10 +160,6 @@ class CommitteeRegressor(EnsembleBase):
             The calibration method to use. Options are "isotonic-regression" or "scaling-factor".
         **kwargs : dict
             Additional keyword arguments to be passed to the committee's predict method.
-
-        Returns
-        -------
-        None
 
         """
         # Validate method selection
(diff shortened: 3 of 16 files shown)


openadmet/models/architecture/chemprop.py
```diff
@@ -22,30 +22,97 @@
...
+    n_tasks: int = 1
+    messages: str = "bond"
+    aggregation: str = "norm"
     depth: int = 3
     message_hidden_dim: int = 300
     ffn_hidden_dim: int = 300
     ffn_num_layers: int = 1
-    messages: str = "bond"
-    aggregation: str = "norm"
+    scaler: object = None
```
hmacdope, line 83: Scaler now as a class attribute seems error prone given previous issues (?) we had with it. I can't remember exactly why this wasn't possible before but IIRC it was because the correct scaler is not available on `__init__`  as dependent on data.

At a minimum this will need changes in anvil + testing.

openadmet/models/architecture/chemprop.py
```diff
@@ -55,18 +122,55 @@ def validate_messages(cls, value):
...
+        Parameters
+        ----------
+        *args : tuple
+            Positional arguments for parent class.
+        **kwargs : dict
+            Keyword arguments for parent class and model configuration.
+
+        """
+        super().__init__(*args, **kwargs)
+        self.build()
```
hmacdope, line 160: Same thing with `Scaler`

openadmet/models/architecture/chemprop.py
```diff
@@ -55,18 +122,55 @@ def validate_messages(cls, value):
...
+
+        Returns
+        -------
+        nn.UnscaleTransform or None
+            The output transform to apply to predictions.
+
         """
-        if scaler is not None:
-            output_transform = nn.UnscaleTransform.from_standard_scaler(scaler)
+        if self.scaler is not None:
```
hmacdope, line 172: Query scaler changes here also (sorry for repetition).

openadmet/models/architecture/chemprop.py
```diff
@@ -79,29 +183,71 @@ def _get_output_transform(self, scaler):
...
+        NotImplementedError
+            Always raised. Use a trainer for training.
+
         """
         raise NotImplementedError(
             "Training not implemented in model class, use a trainer"
         )
 
-    def build(self, scaler=None):
+    def build(self):
```
hmacdope, line 239: Same scaler query

openadmet/models/architecture/chemprop.py
```diff
@@ -154,14 +300,26 @@ def build(self, scaler=None):
...
+                message_passing=mp,
+                agg=aggr,
+                predictor=ffn,
+                batch_norm=self.batch_norm,
+                metrics=metric_list,
+                warmup_epochs=self.warmup_epochs,
+                init_lr=self.init_lr,
+                max_lr=self.max_lr,
+                final_lr=self.final_lr,
+            )
```
hmacdope, line 318: Way nicer

hmacdope (changes requested):
Quick once through, main query is about scaler changes.

openadmet/models/architecture/model_base.py
```diff
@@ -482,3 +437,23 @@ def deserialize(
...
+            Keyword arguments to be passed to the implementing model's `freeze_weights` method.
+
+        Notes
+        -----
+        This method should set the `requires_grad` attribute of the specified layers to False,
+        preventing their weights from being updated during training. It also should set these
+        layers to evaluation mode.
+
+        """
+        logger.warning(f"Weight freezing currently not implemented for {self.type}.")
```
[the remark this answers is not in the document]
smcolby: We can go that route, but since this method gets called well after initialization, the workflow wouldn't fail until pretty late in the game. I'll try to validate early if freezing weights is requested, but model doesn't support.

PR #339, opened 2025-09-18
Remove mod_params and add class level params from MTENN config

### LINKED ISSUE #334: Expose params for MTENN implementation
Currently in architecture/mtenn.py mod_params must be hardcoded. See from_params function

```
    def from_params(self, params):
        pass
```

This should be modified so that we can expose the params for different architectures. 

Disclaimer: When I generalize code for the addition of other MTENN models, this might change the structure of this code. However, for now this is a necessary change.

khuddzu:
## Description
The MTENN anvil architecture was originally coded to only take defaults of the SchNet Representation and for ModelConfig, both classes from the MTENN package. This PR will expose hyper param options to make training a model more flexible. Also, it will satisfy the removal of mod_params, which is a team plan moving forward.  

Will resolve #334 

## Todos
  - [x] Select params to expose and add at class level
  - [x] Rework build function to run with new setup
  - [x] Add docstrings while we are at it
  - [x] Update tests

### DIFF
--- a/openadmet/models/architecture/mtenn.py
+++ b/openadmet/models/architecture/mtenn.py
@@ -5,7 +5,7 @@ from typing import ClassVar
 import torch
 from lightning import pytorch as pl
 from loguru import logger
-from mtenn.config import SchNetRepresentationConfig, ModelConfig
+from mtenn.config import ModelConfig, SchNetRepresentationConfig
 
 from openadmet.models.architecture.model_base import LightningModelBase
 from openadmet.models.architecture.model_base import models as model_registry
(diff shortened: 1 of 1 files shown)


openadmet/models/architecture/mtenn.py
```diff
@@ -106,4 +131,4 @@ def make_new(self) -> "MTENNSchNetModel":
         """
         Copy parameters to a new model instance without copying the estimator
         """
-        return self.__class__(**self.mod_params, **self.dict(exclude={"estimator"}))
+        return self.__class__(**self.dict(exclude={"estimator"}))
```
smcolby, line 134: `dict` method is deprectated, use `model_dump` (it's functionally equivalent). I.e.:

```suggestion
        return self.__class__(**self.model_dump(exclude={"estimator"}))
```

PR #349, opened 2025-09-26
Add jupyter guard to logger rich formatting

hmacdope:
- [X] I certify that this contribution is covered by the MIT License [here](https://github.com/OpenADMET/openadmet_toolkit/blob/main/LICENSE) and the **Developer Certificate of Origin** at <https://developercertificate.org/>.


openadmet/models/log.py
```diff
@@ -2,5 +2,19 @@
 
 from loguru import logger
 from rich.logging import RichHandler
```
khuddzu, line 4: you are double importing this. this is also a criteria in the if is not notebook statement

PR #355, opened 2025-09-30
Update MTENN to include MSE step loss from `drugforge`

### LINKED ISSUE #348: Add step loss to MTENN implementation
For his Moonshot work, Ben used a steploss function. this loss is available in drugforge. there are other loss functions as well they have implemented that we could access.  

Currently I am working on reproducing his results, so I will just hard code step_loss into the mtenn pipeline. However, once this issue is resolved it should be accessible via drugforge (if we are okay with adding that dependency); and maybe we can access different loss functions.

khuddzu:
Adding MSELoss from drugforge, to include step loss functionality. If we are okay with adding drugforge dependency, then I will add using their package. If not, I can hardcode their loss here, which is what I will start with anyways. There are other loss functions as well they have implemented that we could access.

Will close #348 


hmacdope:
@khuddzu hardcoding the loss is preferable, as `drugforge` has an insane amount of dependencies.

hmacdope (changes requested):
Might be worth abstracting this into a `loss.py` setup with some of our custom loss functions.

PR #372, opened 2025-10-03
Feat/369/mtenn predict compat

khuddzu:
## Description
Update the MTENNFeaturizer to allow for prediction. Currently the featurizer requires y as input. The updates allows this variable to be optional. 

- [ ] I certify that this contribution is covered by the MIT License [here](https://github.com/OpenADMET/openadmet_toolkit/blob/main/LICENSE) and the **Developer Certificate of Origin** at <https://developercertificate.org/>.

### DIFF
--- a/openadmet/models/features/mtenn.py
+++ b/openadmet/models/features/mtenn.py
@@ -47,7 +47,7 @@ class MTENNDataset(Dataset):
     def __init__(
         self,
         complexes: Iterable[Path],
-        y: Iterable[Any],
+        y: Iterable[Any] = None,
         ligand_resname: Union[str, list[str]] = "LIG",
         ignore_h: bool = True,
     ):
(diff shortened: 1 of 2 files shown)


openadmet/models/features/mtenn.py
```diff
@@ -266,14 +270,22 @@ def featurize(self, complexes: Iterable[Path], y: Iterable[Any]) -> DataLoader:
...
+                y=y,
+                ligand_resname=self.ligand_resname,
+                ignore_h=self.ignore_h,
+            )
+        else:
+            self._dataset = MTENNDataset(
+                complexes=complexes,
+                ligand_resname=self.ligand_resname,
+                ignore_h=self.ignore_h,
+            )
```
hmacdope, line 288: Because Y is optional and can be none can you not just have a single branch? 

```python
            self._dataset = MTENNDataset(
                complexes=complexes,
                ligand_resname=self.ligand_resname,
                ignore_h=self.ignore_h,
                y=y # None is ok
            )
```

hmacdope (changes requested):
One quick query but looks great otherwise

PR #378, opened 2025-10-07
Add guardrail for multitasking 

### LINKED ISSUE #237: Add guardrails for multitasking
Some workflows allow training against multiple targets, while others do not. 

While the long term goals should be to build a more comprehensive system to check for mis-matches, we should be able to build some simple pydantic guardrails in the meantime. 

Some examples of things that are easy to trip up on 

* `n_tasks` and number of regression targets do not match for ChemProp style models 
* GAT cannot currently do multitasking
* Traditional ML workflows (non-DL) need only a single regression target. 

As well as a few others, this should be easy to solve, but we should keep a vigilant eye on these.

hmacdope:
## Description
Fixes #237 

## Status
- [X] Ready to go

- [X] I certify that this contribution is covered by the MIT License [here](https://github.com/OpenADMET/openadmet_toolkit/blob/main/LICENSE) and the **Developer Certificate of Origin** at <https://developercertificate.org/>.

### DIFF
--- a/openadmet/models/anvil/workflow_base.py
+++ b/openadmet/models/anvil/workflow_base.py
@@ -96,12 +96,8 @@ class AnvilWorkflowBase(BaseModel):
         ValueError
             If the model is multitask but the data specification does not support multitask learning.
         """
-        if len(self.data_spec.target_cols) > 1 and not self.model.n_tasks > 1:
-            raise ValueError(
-                "The model is not multitask but the data specification has multiple target columns."
-            )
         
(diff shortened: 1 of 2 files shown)


hmacdope:
@smcolby could you have a look here?

smcolby:
It looks like `n_tasks` is now getting passed to all of the models that can't handle multitask.

smcolby (changes requested):
Not passing tests for CatBoost, RF, SVM, or Dummy

hmacdope:
This is a bit of a hack but it will do for now.

PR #429, opened 2025-11-06
Change driver to be set from trainer class name rather than from driv…

dwwest:
Setting the driver from the variable in metadata is dangerous, it's not a place users are going to think that any info is being pulled from. I've changed the code so the driver is pulled from the class name of the trainer, so this won't require any anvil recipes to have to be changed.

### DIFF
--- a/openadmet/models/anvil/workflow.py
+++ b/openadmet/models/anvil/workflow.py
@@ -29,7 +29,7 @@ def _safe_to_numpy(X):
 class AnvilWorkflow(AnvilWorkflowBase):
     """Workflow for running basic Anvil configuration."""
 
-    driver: DriverType = DriverType.SKLEARN
+    _driver_type: DriverType = DriverType.SKLEARN
 
     @model_validator(mode="after")
     def check_if_val_needed(self):
@@ -232,7 +232,7 @@ class AnvilWorkflow(AnvilWorkflowBase):
         logger.info(f"Running workflow from directory {output_dir}")
 
         # Log workflow driver selection
-        logger.info(f"Running with driver {self.driver}")
+        logger.info(f"Running with driver {self._driver_type}")
 
         # Load data from YAML specification
         logger.info("Loading data")
@@ -406,7 +406,7 @@ class AnvilWorkflow(AnvilWorkflowBase):
 class AnvilDeepLearningWorkflow(AnvilWorkflowBase):
     """Workflow for running deep learning Anvil configuration."""
 
-    driver: DriverType = DriverType.LIGHTNING
+    _driver_type: DriverType = DriverType.LIGHTNING
 
     @model_validator(mode="after")
     def check_no_transform(self):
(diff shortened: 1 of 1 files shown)


openadmet/models/anvil/workflow.py
```diff
@@ -824,3 +824,32 @@ def run(
...
+        If trainer type is unrecognized.
+
+    """
+    trainer_lower = trainer.lower()
+    if "sklearn" in trainer_lower:
+        return Drivers.SKLEARN
+    elif "pytorch" in trainer_lower or "lightning" in trainer_lower:
+        return Drivers.PYTORCH
+    else:
+        raise ValueError(f"Unrecognized trainer type: {trainer}")
```
smcolby: I still think this should be an attribute rather than inferred from the class name.

openadmet/models/anvil/specification.py
```diff
@@ -627,6 +627,9 @@ class ProcedureSpec(SpecBase):
     ensemble: EnsembleSpec | None = None
     train: TrainerSpec
     transform: Optional[TransformSpec] = None  # Optional transform step
+    driver: Optional[str] = (
+        None  # Optional driver override, otherwise inferred from trainer
+    )
```
smcolby: I don't think we should allow this

openadmet/models/anvil/specification.py
```diff
@@ -393,7 +393,7 @@ class Metadata(SpecBase):
         ..., description="The version of the metadata schema."
     )
     driver: str = Field(
```
hmacdope: Not you but slight improvement

```suggestion
    driver: DriverType = Field(
```
And then the next line can be 
`DriverType.SKLEARN`

hmacdope (changes requested):
This is positive change, thanks @dwwest. Just two small nits.

openadmet/models/anvil/specification.py
dwwest: I think we need to leave this the way it is, because otherwise we break backwards compatibility? There were two specs for drivers and one of them used "pytorch" and one of them used "lightning." The one reading from metadata expects "pytorch," but I went with the one that uses "lightning." Lmk if we prefer one name over the other?

smcolby (changes requested):
Soooo close!

openadmet/models/anvil/workflow.py
```diff
@@ -29,7 +29,7 @@ def _safe_to_numpy(X):
 class AnvilWorkflow(AnvilWorkflowBase):
     """Workflow for running basic Anvil configuration."""
 
-    driver: Drivers = Drivers.SKLEARN
+    driver: DriverType = DriverType.SKLEARN
```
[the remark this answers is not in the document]
dwwest: Ok, I changed the name, I think the driver just needs the attribute for the log statement.

hmacdope (approved):
LGTM from my perspective

PR #438, opened 2025-11-19
[WIP] Add clustering method to cross validation

### LINKED ISSUE #430: Implement clustering splitter in models

dwwest:
This PR has two purposes: 

1) Removing stratify from the clustering splitter under split/ in the main repo
2) Adding a temporary way to do cluster splitting in cross validation within eval/ with the eventual idea being to merge the splitter under split/ into cv


openadmet/models/eval/cross_validation.py
```diff
@@ -61,6 +67,10 @@ class CrossValidationBase(EvalBase):
         Minimum value for the axes.
     max_val : float
         Maximum value for the axes.
+    use_clustering : bool
+        Whether to use clustering for splitting the data.
+    clustering_method : str
+        Method to use for clustering ('butina', 'kmeans', 'bemis-murcko').
```
smcolby: These args on CV base still seems odd to me. "Base" should be base functionality, then subclass to add functionality. Particularly since KFoldRepeatedCrossValidation won't use that flag

openadmet/models/eval/cross_validation.py
```diff
@@ -375,6 +451,154 @@ def write_report(self, output_dir):
...
+
+    def evaluate(
+        self,
+        model=None,
+        X_train=None,
+        y_train=None,
+        y_pred=None,
+        y_true=None,
+        X_all=None,
+        X_smiles_all=None,
```
smcolby: These two were added by you? Remind me what for?

smcolby (commented):
Getting there! Some interim feedback

hmacdope (commented):
Can the splitter be passed in as an argument instead of repeating the whole implementation or is it just too complicated?

PR #444, opened 2025-12-03
Rework splitters to return group id for use in cross-validation

dwwest:
This PR will:

1) Rework existing splitters so that they return a group ID for cross-validation in addition to the train, test, and val
2) Add a cluster splitter with the same outputs
3) Add group-based splitting into cross-validation as in branch remove-stratify
4) Add any passing of necessary parameters in workflow.py, while maintaining backwards compatibility with old anvil structure

### DIFF
--- a/openadmet/models/split/cluster.py
+++ b/openadmet/models/split/cluster.py
@@ -95,7 +95,7 @@ class ClusterSplitter(SplitterBase):
             transformer = MoleculeTransformer(
                 vec_featurizer,
                 parallel_kwargs={"progress": False},
-                )
+            )
             with dm.without_rdkit_log():
                 feat, _ = transformer(X, ignore_errors=True)
             fp_list = list(np.squeeze(feat))
@@ -154,9 +154,14 @@ class ClusterSplitter(SplitterBase):
             subarrays_y, [train_idxs, val_idxs, test_idxs]
         )
 
-        if self.train_size != len(X_train) or self.val_size != len(X_val) / total_elements or self.test_size != len(X_test) / total_elements:
+        if (
+            self.train_size != len(X_train)
+            or self.val_size != len(X_val) / total_elements
+            or self.test_size != len(X_test) / total_elements
+        ):
             logging.warning(
-                f"Train/val/test sizes DO NOT match input requests due to cluster sizes: Train: {self.train_size/total_elements}, Val: {self.val_size/total_elements}, Test: {self.test_size/total_elements}")
+                f"Train/val/test sizes DO NOT match input requests due to cluster sizes: Train: {self.train_size / total_elements}, Val: {self.val_size / total_elements}, Test: {self.test_size / total_elements}"
+            )
 
         # Return train, val and test sets
         return X_train, X_val, X_test, y_train, y_val, y_test, clusters
--- a/openadmet/models/split/scaffold.py
+++ b/openadmet/models/split/scaffold.py
@@ -228,7 +228,9 @@ class MaxDissimilaritySplitter(SplitterBase):
             - y_test: Test set target values (or None if test_size=0).
 
         """
-        logging.warning("MaxDissimilaritySplitter is not available for cross-validation.")
+        logging.warning(
+            "MaxDissimilaritySplitter is not available for cross-validation."
+        )
         # No test set requested
         if self.test_size == 0:
             # Split into train and val


openadmet/models/split/cluster.py
```diff
@@ -69,58 +75,88 @@ def split(self, X, y):
...
-        if self.test_size == 0:
-            # Split into train and val
-            X_train, X_val, y_train, y_val = train_test_split(
-                X,
-                y,
-                train_size=None,
-                test_size=int(self.val_size * X.shape[0]),
+            logging.warning(
+                "KMeans clustering is NOT DETERMINISTIC even with random seed."
+            )
```
smcolby: Is this true of inter-platform (i.e. mac versus linux), or intra-platform as well (i.e. repeat mac runs)?

openadmet/models/split/cluster.py
```diff
@@ -69,58 +75,88 @@ def split(self, X, y):
...
+            km = KMeans(
+                n_clusters=self.k_clusters,
+                n_init=1,
                 random_state=self.random_state,
-                stratify=clusters,
+                algorithm="lloyd",
+            )
+            fp_list = [smi2numpy_fp(x).astype(np.float64) for x in X]
+            with threadpool_limits(limits=1):
```
smcolby: I'm guessing this was an attempt to fix the reproducibility issues between platforms? Do we still need it if we've accepted that fate?

openadmet/models/split/cluster.py
```diff
@@ -69,58 +75,88 @@ def split(self, X, y):
...
+            cum_counts = np.cumsum(shuffled_lengths)
+
+            # Searchsorted finds the first index where cum_counts >= target
+            split_1 = np.searchsorted(cum_counts, target_counts[0])
+            split_2 = np.searchsorted(cum_counts, target_counts[1])
+
+            # Look at how far the actual cut points are from ideal targets
+            error = abs(cum_counts[split_1] - target_counts[0]) + abs(
+                cum_counts[split_2] - target_counts[1]
             )
```
smcolby: I'm guessing this is only done for 2 of 3 splits, since the third split will just be "the rest"?

openadmet/models/split/cluster.py
```diff
@@ -69,58 +75,88 @@ def split(self, X, y):
...
-            stratify=split_clusters,
+        logging.warning(val_idxs)
+
+        # Retrieve train, val, and test sets for X and y separately
+        X_train, X_val, X_test = retrieve_data_by_idx(
+            subarrays_X, [train_idxs, val_idxs, test_idxs]
+        )
+        y_train, y_val, y_test = retrieve_data_by_idx(
+            subarrays_y, [train_idxs, val_idxs, test_idxs]
         )
```
smcolby: Are these effectively the same as:
```python
X_train = subarrays_X[train_idxs]
X_val = subarrays_X[val_idxs]
X_test = subarrays_X[test_idxs]
```
With special handling for cases when index is empty (e.g. if no val set requested)?

smcolby: I suppose I should just look at `retrieve_data_by_idx`...

smcolby: Yep, that's exactly how it works. Plus it concatenates the subarrays. Disregard this thread.

openadmet/models/split/scaffold.py
```diff
@@ -34,6 +39,7 @@ def split(self, X, y):
             - y_test: Test set target values (or None if test_size=0).
 
         """
+        logging.warning("ScaffoldSplitter is not available for Cross-Validation.")
```
smcolby: Why not? Also replace "Cross-Validation" with "cross-validation" IMO

openadmet/models/split/scaffold.py
```diff
@@ -211,6 +228,9 @@ def split(self, X, y):
             - y_test: Test set target values (or None if test_size=0).
 
         """
+        logging.warning(
+            "MaxDissimilaritySplitter is not available for Cross-Validation."
+        )
```
smcolby: Same here: why not? Don't these also just return groups, which can then go into the CV functionality? 

"Cross-Validation" -> "cross-validation"

smcolby (changes requested):
So far so good!

openadmet/models/split/cluster.py
dwwest: Ah, great point, just inter-platform

openadmet/models/split/scaffold.py
dwwest: They don't return groups, which is the issue. I'd have to go into the splito code to figure out how to pull that out, and it's different for all three. I can fix this, but I want to punt to a later PR.

openadmet/models/split/cluster.py
```diff
@@ -69,58 +75,88 @@ def split(self, X, y):
...
+                "KMeans clustering is NOT DETERMINISTIC even with random seed."
+            )
+            km = KMeans(
+                n_clusters=self.k_clusters,
+                n_init=1,
                 random_state=self.random_state,
-                stratify=clusters,
+                algorithm="lloyd",
+            )
+            fp_list = [smi2numpy_fp(x).astype(np.float64) for x in X]
```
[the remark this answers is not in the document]
dwwest: I agree, but it needs to be done again for type/compatibility reasons with pretty much every off-the-shelf clustering method

openadmet/models/split/cluster.py
```diff
@@ -69,58 +75,88 @@ def split(self, X, y):
...
-            split_clusters = get_bemis_murcko_clusters(X_train_val)
-        elif self.method == "kmeans":
-            split_clusters = get_kmeans_clusters(
-                X_train_val, n_clusters=self.k_clusters
-            )
+        best_indices, s1, s2 = best_split
+
+        train_idxs = best_indices[: s1 + 1]
+        val_idxs = best_indices[s1 + 1 : s2 + 1]
+        test_idxs = best_indices[s2 + 1 :]
```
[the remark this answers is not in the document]
dwwest: I do think leaving it the way it is for now is good, because it is less code and the setup now is pretty specific to picking the right inds for cluster splitting?

PR #451, opened 2025-12-09
Update inference to allow for parquet input path for predict

khuddzu:
`predict` only allowed for CSV or SDF as the input file for inference. We want to allow for more file types. This fix adds the ability to input a file ending in ".parquet".


hmacdope (approved):
LGTM, pending CI issues

PR #468, opened 2026-01-16
Add ability to load in pretraining weights for finetuning in chemprop

dwwest:
We want to be able to load weights from pretraining from external sources.

This PR will affect only chemprop, but maybe this could eventually be extended to other foundation models.


hmacdope:
Closing here as not under development. @dwwest let me know if incorrect

PR #505, opened 2026-03-04
Fix pairwise inference task index selection

SergeiNikolenko:
## Description
Fix pairwise inference dataframe generation to use explicit task index selection (`task_idx`) and add regression tests for single-task and multitask pairwise paths.

## Status
- [x] Ready to go

### DIFF
--- a/openadmet/models/inference/inference.py
+++ b/openadmet/models/inference/inference.py
@@ -93,30 +93,35 @@ def load_anvil_model_and_metadata(model_dir):
 
 
 def _generate_pairwise_df(
-    data, input_col, feat, predictions, predictions_tag, std_tag
+    data,
+    input_col,
+    feat,
+    predictions,
+    std,
+    predictions_tag,
+    std_tag,
+    task_idx=0,
+    pairwise_df=None,
 ) -> pd.DataFrame:
     """Generate a DataFrame for pairwise predictions."""
-    smiles = data[input_col].values
-    pairwise_dataset = PairwiseAugmentedDataset(smiles, None, how=feat.how_to_pair)
-    pairs = pairwise_dataset.idxs  # list of (i, j) tuples
+    if pairwise_df is None:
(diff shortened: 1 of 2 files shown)


smcolby (changes requested):
The fix to `inference.py` is correct and the core test (`test_generate_pairwise_df_uses_task_idx_column`) is solid. A few changes are required before merge.
 
 **1. Test style: replace `monkeypatch` + hand-rolled dummies with `pytest-mock`**
 
 `test_predict_single_task_pairwise_uses_column_zero` and `test_predict_pairwise_multitask_passes_task_idx` both use `monkeypatch.setattr` with custom nested dummy classes (`DummyPairwiseFeaturizer`, `DummyModel`, `fake_loader`, `fake_generate_pairwise_df`). We're trying to use `pytest-mock` with `mocker.patch(..., autospec=True)` for this sort of thing, and will be updating the contribution guidelines soon.
 
 Please rewrite both tests using the `mocker` fixture, e.g.:
 
 ```python
 def test_predict_single_task_pairwise_uses_column_zero(mocker, tmp_path):
     mock_load = mocker.patch(
         "openadmet.models.inference.inference.load_anvil_model_and_metadata",
         autospec=True,
     )
     mock_load.return_value = (...)
     ...
 ```
 
 `test_generate_pairwise_df_uses_task_idx_column` is fine as-is — keep it.
 
 **2. `DummyPairwiseFeaturizer` is copy-pasted verbatim** between the two`monkeypatch` tests. Once rewritten with `mocker`, this duplication goes away naturally.

openadmet/models/inference/inference.py
dwwest, line 266: Let's change j to task_idx for consistency.

dwwest (changes requested):
This looks good, thank you for catching this bug. Just a quick cosmetic change, plus Sean's requests re: testing, and it should be good to go.

Issue #509, opened 2026-03-04
[CHORE] Ensure shape is handled consistently across package


smcolby:
Should confirm any and all instances in which input shapes are checked for and coerced, i.e. to handle (N,), (N, 1), and (N, M) cases when input to downstream functions may be opinionated in different ways (i.e. accepts (N,) but not (N, 1)), are handled in a functionally similar way. Does not have to be _literally_ the same way, but all patterns should result in the same outcome.

## 1. `openadmet/models/eval/regression.py` — Lines 99–102 & 383–386

**Pattern:** `(N,)` → `(N, 1)` for both `y_pred` and `y_true`.

```python
if y_pred.ndim == 1:
    y_pred = y_pred.reshape(-1, 1)
if y_true.ndim == 1:
    y_true = y_true.reshape(-1, 1)
```

**What it does:** Appears in both `RegressionMetrics.evaluate()` and `RegressionPlotter.evaluate()`. Coerces any flat `(N,)` array into `(N, 1)` so that `shape[1]` can be read as the number of tasks. The downstream per-task loop over `y_true[:, task_id]` requires 2D arrays. Does **not** handle the extra case where `y_true` is `(N, 1)` but `y_pred` is `(N,)` — those are handled independently for each argument. No coercion is applied for `(N, M)` beyond reading `shape[1]`.

## 2. `openadmet/models/eval/uncertainty.py` — Lines 125–130 & 325–330

**Pattern:** `(N,)` → `(N, 1)` for `y_pred`, `y_true`, and `y_std`.

```python
if y_pred.ndim == 1:
    y_pred = y_pred.reshape(-1, 1)
if y_true.ndim == 1:
    y_true = y_true.reshape(-1, 1)
if y_std.ndim == 1:
    y_std = y_std.reshape(-1, 1)
```

**What it does:** Appears in both `UncertaintyMetrics.evaluate()` and `UncertaintyPlotter.evaluate()`. Same pattern as regression — coerces all three arrays to `(N, 1)` before iterating over tasks. After coercion, each task column is extracted and immediately `.flatten()`'d back to `(N,)` before passing to metric functions (lines 146–148, 346–348).

## 3. `openadmet/models/eval/classification.py` — Lines 114, 117–135, 324–329, 374–382

**Pattern:** Detects `(N,)` or `(N, 1)` as the binary case; `(N, K)` with `K > 1` as the multiclass case.

```python
[...]

PR #512, opened 2026-03-05
Make shape handling consistent across package for #509

dwwest:
See #509


openadmet/models/eval/regression.py
```diff
@@ -174,10 +174,8 @@ def evaluate(
             y_true = y_true.to_numpy()
 
         # Ensure y_pred and y_true are 2D arrays for consistency
-        if y_pred.ndim == 1:
-            y_pred = y_pred.reshape(-1, 1)
-        if y_true.ndim == 1:
-            y_true = y_true.reshape(-1, 1)
+        y_pred = ensure_2d(y_pred)
+        y_true = ensure_2d(y_true)
```
smcolby, line 178: Oh this is slick. So just need to see if this function is rock solid and we're good to go!

openadmet/models/eval/utils.py
```diff
@@ -1,5 +1,27 @@
...
+    Returns
+    -------
+    np.ndarray
+        Array of shape (N, 1) if input was 1D, otherwise the original array.
+
+    """
+    if arr.ndim == 1:
+        return arr.reshape(-1, 1)
+    return arr
+
```
smcolby, line 24: Just thinking of edge cases... This only does something if the array is flat: (N, ) -> (N, 1). If it's already 2D (i.e. (N, M)) we're good. If it's more than 2D (N, M, K), also good, and no reason the latter would need coercion. 

LGTM!

smcolby (changes requested):
Can you do one last sanity check that this is ALL cases where shapes may need coercion to 2D? Seems sus we only had to touch 2 files (regression.py and uncertainty.py).

smcolby (approved):
LGTM

PR #532, opened 2026-04-20
Add return_members option to CommitteeRegressor.predict()

### LINKED ISSUE #464: [ENH] When predicting with an ensemble there should be an option to return the predictions of each ensemble member
We should have an option to return individual ensemble member predictions.

hmacdope:
Closes #464

## Summary

- Adds `return_members=False` to `CommitteeRegressor.predict()` and `_predict()`
- When `True`, the raw per-member predictions are appended to the return tuple with shape `(n_samples, n_tasks, n_members)`
- Composes with `return_std` — all four combinations are supported:
  - `predict(X)` → `mean`
  - `predict(X, return_std=True)` → `(mean, std)`
  - `predict(X, return_members=True)` → `(mean, members)`
  - `predict(X, return_std=True, return_members=True)` → `(mean, std, members)`
- No changes to existing call sites — fully backwards compatible

## Test plan

- [ ] `test_return_members` unit test covers shape, consistency with mean/std, and the no-flag scalar return

🤖 Generated with [Claude Code](https://claude.com/claude-code)


hmacdope:
@smcolby can you review here?

smcolby (changes requested):
A bit of pedantry: `return_members` gives the impression that we'd be returning the members themselves. What do you thinking about `return_all`?

Second, when returning the full set of predictions, do we want to return _only_ the `(n_samples, n_tasks, n_members)`-shaped array, and not the mean and std? I.e. one would assume a user would be using the full array downstream, so mean and std calcs would be handled as needed. Of course this would require some argument validation to make `return_std` and `return_all` (or whatever we decide) mutually exclusive.

smcolby (approved):
Wonderful!

PR #548, opened 2026-05-14
Fix pandas 3.0 compatibility in PostHocComparison

smcolby:
## Description

pandas 3.0 raises TypeError when assigning strings ('***', '**', '*', '') into a float64 DataFrame column. Cast significance DataFrame to object dtype before masked string assignment.

pandas 3.0 also preserves dict keys as the row index when a dict is assigned to a column. levene_test built each column as {'stat': ..., 'pvalue': ...}, so the index changed from [0, 1] to ['stat', 'pvalue']. Update tests to use explicit label-based access (levene[metric]['stat']) rather than positional.

## Status

---

### DIFF
--- a/openadmet/models/tests/unit/comparison/test_comparison.py
+++ b/openadmet/models/tests/unit/comparison/test_comparison.py
@@ -81,8 +81,8 @@ def test_posthoc_comparison():
     levene, tukeys_df = comp_obj.compare(
         model_stats_fns=model_stats, labels=model_tags, task_names=task_tags
     )
-    assert_almost_equal(levene["mse"][0], 1.2975061710820235)
-    assert_almost_equal(levene["ktau"][0], 0.8835355632672074)
+    assert_almost_equal(levene["mse"]["stat"], 1.2975061710820235)
+    assert_almost_equal(levene["ktau"]["stat"], 0.8835355632672074)
     assert_almost_equal(tukeys_df["metric_val"][0], 0.10937620875054632)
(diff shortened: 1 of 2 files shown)


openadmet/models/tests/unit/comparison/test_comparison.py
```diff
@@ -81,8 +81,8 @@ def test_posthoc_comparison():
     levene, tukeys_df = comp_obj.compare(
         model_stats_fns=model_stats, labels=model_tags, task_names=task_tags
     )
-    assert_almost_equal(levene["mse"][0], 1.2975061710820235)
-    assert_almost_equal(levene["ktau"][0], 0.8835355632672074)
+    assert_almost_equal(levene["mse"]["stat"], 1.2975061710820235)
+    assert_almost_equal(levene["ktau"]["stat"], 0.8835355632672074)
     assert_almost_equal(tukeys_df["metric_val"][0], 0.10937620875054632)
```
khuddzu, line 86: why does tukeys_df still use indices instead of object?

khuddzu (approved):

openadmet/models/tests/unit/comparison/test_comparison.py
smcolby: tukeys_df rows are numbered 0, 1, 2… so integer access is correct. levene rows are named "stat" and "pvalue" (only 2 of them) and were previously being accessed by [0, 1], which is no longer allowed in pandas 3.

PR #568, opened 2026-06-11
Adopt AGENTS.md and repo-local Claude rules

smcolby:
## What

Replaces the single, stale `.github/copilot-instructions.md` with a canonical, harness-agnostic `AGENTS.md`, and deploys scoped coding rules under `.claude/rules/` that activate by file pattern.

Opened as a **draft** for review; the instruction content was minted fresh against the current codebase rather than carried over from the old doc.

## Why

The previous `.github/copilot-instructions.md` had drifted from the code:

- It described `registries.py` as wildcard imports with load-bearing import order. The package actually uses a lazy loader (`_registry_loader.py`) with decorator-based registration.
- It listed four registries; eight `ClassRegistry` objects exist.
- Its install command (`pip install -e --no-deps .`) is malformed.

## What's here

- **`AGENTS.md`** (new canonical instruction file): environment (conda/mamba), commands, the registry + lazy-loader architecture, the Pydantic ABC component contract, the two model families, and conventions.
- **`.claude/rules/`**: path-scoped rules that Claude Code activates when matching files are read.
  - Provenance-stamped renders of shared catalog rules that apply here: `python-docs`, `python-testing`, `machine-learning`, `medicinal-chemistry`.
  - An adapted `python-core` (marked `diverged`) reconciled to this repo's Python 3.10 floor, ABC-based component model, and actual ruff-`D`/black/isort/flake8 gate.
  - A bespoke `openadmet-models` rule: the registry contract, Lightning-only training, and this repo's splitter-exclusivity and save/load round-trip test discipline.
- **`CLAUDE.md`**: imports `AGENTS.md`.
- **`.github/copilot-instructions.md`**: shrunk to a pointer at `AGENTS.md`.

## Notes

- No source or test code changes; instructions and rules only.
- The rules are guidance for agents, not new gate enforcement; the pre-commit gate is unchanged.


hmacdope:
@smcolby I think this might belong at root rather than in this repo

Issue #571, opened 2026-06-16
Make shuffle and drop_last training-only across all deep-learning workflows


smcolby:
## Summary

Shuffling and `drop_last` are currently properties of the *featurizer instance*, but a single featurizer builds the train, validation, test, **and** inference loaders. Because predictions are concatenated in dataloader-iteration order and paired positionally with `y_true`, any shuffle or dropped batch on a non-training loader silently corrupts metrics (or crashes). This issue proposes a single, unified fix that makes shuffle/`drop_last` *training-only* across every workflow, so the safe behavior is the default and the dangerous behavior cannot be reached by accident.

This unifies and supersedes the two narrower reports:
- #383 — setting `shuffle=True` breaks workflow prediction and evaluation
- #517 — `drop_last` in `ChemPropFeaturizer` causes inference crash and silent metric corruption

## Root cause

The same featurizer (or a `make_new()` copy that propagates the flags) builds every loader, and `shuffle`/`drop_last` are read from instance state with no per-call control:

- `ChemPropFeaturizer.featurize` builds loaders from `self.shuffle` (`features/chemprop.py:170-175`); `_vendor_build_dataloader` sets `drop_last=True` whenever `len(dataset) % batch_size == 1` (`features/chemprop.py:78-81`), regardless of split role.
- `PairwiseFeaturizer.featurize` likewise builds from `self.shuffle` (`features/pairwise.py:175-180`).
- Predictions return in loader order with no index bookkeeping: `ChemPropModel.predict` does `torch.cat(trainer.predict(estimator, loader)).numpy()` (`architecture/chemprop.py:654-655`); `NepareModel.predict` is the same shape (`architecture/nepare.py:272`).
- Evaluators pair positionally: held-out test pairs `y_pred` against `y_test` (`anvil/workflow.py:887-909`); CV pairs `y_pred_fold` against `y_val` after `featurizer.make_new()` copies `shuffle` (`eval/cross_validation.py:663-702`).

Consequences when `shuffle: true` is set in a `feat:` block:
1. **Shuffle (#383):** test/val/CV loaders permute, so `y_pred[i]` no longer matches `y_true[i]`. No crash; metrics are plausible but wrong.
2. **drop_last (#517):** when `len(dataset) % batch_size == 1`, the eval/inference loader returns one fewer row than `y_true`, giving an off-by-one across the whole vector or a shape mismatch.

[...]

PR #572, opened 2026-06-16
Make shuffle and drop_last training-only across deep-learning workflows

### LINKED ISSUE #383: Setting `shuffle=True` in deep learning featurization will break workflow prediction and evaluation
Modifying the `shuffle` argument in deep learning featurizers (thereby configuring the resultant data loaders) will cause shuffled predictions. When then compared to `y_true`, the out-of-order result will be completely wrong. 

Ideally we should be able to choose whether to shuffle training data batches, but force no shuffling for validation and test. Currently there is only a single toggle to affect all splits.

For now, DO NOT SET `shuffle=True` per the issue with test set shuffling.

smcolby:
## What

Makes `shuffle` and `drop_last` **training-only** in the deep-learning featurizers, so validation, test, cross-validation, and inference loaders always preserve input order and length.

Implements the plan in #571 and closes the two underlying bugs:
- Closes #383 — `shuffle=True` permuted eval loaders, silently corrupting metrics (predictions are concatenated in loader order and paired positionally with `y_true`).
- Closes #517 — the size-1 `drop_last` guard dropped a test/inference row, misaligning the whole vector or crashing.

## How

- Add a `train: bool = False` parameter to the `DeepLearningFeaturizer.featurize` contract, honored by `ChemPropFeaturizer` and `PairwiseFeaturizer`.
- `train=True` → `shuffle=self.shuffle` and the batch-norm `drop_last` guard apply. `train=False` (default) → `shuffle=False`, `drop_last=False`, unconditionally.
- Thread `train=True` through the three training call sites: the deep-learning workflow train loader, the ensemble bootstrap train loader (`anvil/workflow.py`), and the CV fold train loader (`eval/cross_validation.py`). All eval/inference loaders keep the safe default.
- Correct the inverted sampler guard in `_vendor_build_dataloader` (`if sampler is not None` → `if sampler is None`), which made the class-balance and seeded-sampler paths unreachable. Inert for the current workflow path (no `seed`/`class_balance` passed), so no behavior change there.
- The inference point-patch (`feat.shuffle = False`) is retained as documented defense-in-depth.

## Contract

| Loader | shuffle | drop_last |
|---|---|---|
| train (`train=True`) | follows `feat.shuffle` | drops a size-1 final batch (batch-norm guard) |
| val / test / CV-val / inference (`train=False`) | forced off | forced off (every row returned) |

## Tests

Added four regression tests in `tests/unit/features/test_features.py` (two per featurizer), asserting on the real `DataLoader`:
- eval loader (`train=False`) with `shuffle=True` → `SequentialSampler`, `drop_last is False`, all rows returned (fails on old code: #383 + #517).
- train loader (`train=True`) → `RandomSampler`, drops the size-1 final batch.

Full unit suite for `features`, `inference`, `anvil`, and `eval` passes; pre-commit clean.

## Status

---

### DIFF
--- a/openadmet/models/anvil/workflow.py
+++ b/openadmet/models/anvil/workflow.py
@@ -614,6 +614,7 @@ class AnvilDeepLearningWorkflow(AnvilWorkflowBase):
                 self.feat.featurize(
                     X_train_bootstrap,
                     y_train_bootstrap,
+                    train=True,
                 )
             )
 
@@ -786,6 +787,7 @@ class AnvilDeepLearningWorkflow(AnvilWorkflowBase):
         train_dataloader, _, train_scaler, train_dataset = self.feat.featurize(
             X_train,
             y_train,
+            train=True,
         )
         torch.save(train_dataloader, output_dir / "train_dataloader.pth")
 
--- a/openadmet/models/eval/cross_validation.py
+++ b/openadmet/models/eval/cross_validation.py
@@ -663,7 +663,7 @@ class PytorchLightningRepeatedKFoldCrossValidation(CrossValidationBase):
             fold_featurizer = featurizer.make_new()
 
             fold_train_dataloader, _, fold_train_scaler, _ = fold_featurizer.featurize(
-                X_train, y_train
+                X_train, y_train, train=True
             )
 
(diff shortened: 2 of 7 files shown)


hmacdope (approved):
LGTM

Issue #499, opened 2026-02-28
[ENH] Inconsistent random seed handling


smcolby:
## Description
We are explicitly controlling seed for splitting, and in some instances model state (where the underlying model has a `random_state` field), but finer-tuned, consistent control would be beneficial, particularly in seeding initialization of deep learning models and in handling bootstrap-resampled data when `use_bagging=True`. What's there now *works*, and isn't necessarily causing issues, just could be better implemented.

### 1. Inconsistent Seed Handling
- Overloaded Seed Source: Both workflow types derive the model/ensemble seed from self.split.random_state. This conflates data splitting reproducibility with model initialization reproducibility. Changing the data split seed shouldn't necessarily force a change in model initialization seeds, and vice-versa.
- Single-Model vs. Ensemble:
    - Ensemble: Logic exists to seed each member (global_seed + i).
    - Single-Model: There is no explicit seeding in the _train methods for single models. Reproducibility relies entirely on external configuration or defaults.

### 2. Model Type Differences and Gaps
- Classic ML (AnvilWorkflow):
    - Relies on checking hasattr(model, "random_state"). If a model doesn't have this attribute, the seed is ignored with a warning, leading to uncontrolled behavior.
    - There is no fallback mechanism for models that might accept a seed via a different argument or global state.
- Deep Learning (AnvilDeepLearningWorkflow):
    - Scratch Training: Correctly calls pl.seed_everything(global_seed + i).
    - Finetuning/Deserialization: Completely omits pl.seed_everything. This means finetuning runs are not guaranteed to be reproducible even if the split seed is fixed.

### 3. Bootstrapping Issues
- Global State Mutation: The bootstrapping logic calls np.random.seed(...) inside the loop. This mutates the global NumPy random state, which can have side effects on other parts of the program or libraries running in the same process.
- Recommendation: Use a local numpy.random.Generator instance (e.g., rng = np.random.default_rng(seed)) for sampling indices.

### 4. Unaccounted Paths
- DL Finetuning: As noted above, the deserialize path in DL workflows lacks seeding.
- Featurization: In the DL ensemble loop, featurizers are re-created, but if they have any stochastic components, they are not explicitly seeded before creation in all paths.

smcolby:
Cross-referencing from the shuffle/drop_last work in #383 / #517 (fix in #572): an adversarial review of that PR surfaced a seeding gap that belongs to this issue.

[...]

PR #573, opened 2026-06-16
Consolidate random seeding under random_seed (global + per-section)

smcolby:
## What

Consolidates random-seed handling under a single `random_seed` interface with layered global + per-section control. Implements the direction in #499 and addresses the unseeded-shuffle gap surfaced by the #572 review.

**Stacked on #572** (`fix/shuffle-droplast-training-only`) — the featurizer seed wiring builds on that PR's `train` flag. Review/merge #572 first; this PR's own diff is the commits after it.

## Naming and the deprecated alias

- Every component exposes one field: **`random_seed`**, and all internal logic uses only that name.
- The legacy **`random_state`** is mapped onto `random_seed` (with one `DeprecationWarning`) in a **single before-validator on `AnvilSection`**, at the config boundary. There is no per-component alias handling and no `_seed` module/mixin. Configs and serialized recipes using `random_state` keep loading.
- Trade-off (by design): the alias is handled for config/spec-driven construction. Direct programmatic `RFRegressorModel(random_state=...)` is not mapped — use `random_seed` in code.
- Each component still hands its underlying library that library's native seed argument: sklearn-style estimators get `random_state` via `seed_to_sklearn_kwargs` (now in `model_base`); XGBoost remaps to `random_state`; CatBoost's native arg is already `random_seed`.

## Layered resolution

Precedence, applied in `AnvilSpecification.to_workflow`:

```
per-section seed (section params)  >  procedure-level global random_seed
```

- `ProcedureSpec.random_seed` defaults to **42** (a real global default). It fills any section that did not set its own seed; an explicit per-section seed always wins.
- **Model-init seeding is decoupled from the split seed** (previously `global_seed = self.split.random_state`), resolved onto `AnvilWorkflowBase.random_seed`.

## Seeding behavior (#499)

- **Single-model and finetune/deserialize paths now seed** via `pl.seed_everything` — previously only the ensemble path did.
- **Featurizer training shuffle is reproducible:** `ChemPropFeaturizer` threads the seed into a `SeededSampler`; `PairwiseFeaturizer` uses a seeded `torch.Generator`. Eval/inference loaders stay unshuffled (from #572).
- **Bootstrap resampling uses a local `np.random.default_rng`** instead of mutating global NumPy state, and is driven by a new optional **`EnsembleSpec.bootstrap_seed`** (falls back to `random_seed`) so data resampling and model init can be seeded independently. Model init uses `self.random_seed + i`; bootstrap uses `bootstrap_seed + i`.
- **Active-learning committee** varies members by `random_seed + i` (keyed on the one name), so members are no longer identical.

## Renames

`random_state` → `random_seed` on every component that exposes a seed: splitters (`SplitterBase`), RF / LGBM / SVM / Dummy / TabPFN, the impute transform, the CV evaluators, and the XGBoost/CatBoost passthrough models. There are no remaining `random_state`-only models.

## PairwiseFeaturizer base change

`PairwiseFeaturizer` moved from `FeaturizerBase` to `DeepLearningFeaturizer`. Rationale: it returns a `DataLoader` and needs the `train` kwarg (#572) and `random_seed` — exactly the `DeepLearningFeaturizer` contract. It was previously duplicating that contract on `FeaturizerBase`; the move inherits the shared `featurize(train=...)` signature and `random_seed` field. It is semantically a deep-learning featurizer (builds tensor batches for the NEPARE torch model).

## Tests

- `tests/unit/test_seed.py`: section-boundary alias maps `random_state` → `random_seed` with a `DeprecationWarning`; direct `random_seed` warns nothing; explicit per-section seed wins over the alias; `seed_to_sklearn_kwargs` rename; resolution precedence helpers.
- `tests/unit/features/test_features.py`: seeded training shuffle is reproducible across global RNG state and varies with the seed.
- Existing fixtures/configs migrated to `random_seed`. Full unit suite: 208 passed, 3 skipped; pre-commit clean.

## Intended behavior change

Decoupling model-init from the split seed changes ensemble member seeds for configs that previously relied on `split.random_state` driving model init. A config that seeds only the split no longer seeds the model — set the procedure-level global `random_seed` (default 42) or a per-section model seed. The companion optimus-prime recipes set the global.

Refs #499.


openadmet/models/active_learning/committee.py
```diff
@@ -293,13 +293,10 @@ def train(
...
-            # Update random state if present
+            # Vary each member's seed so the ensemble is not a set of identical models
             current_mod_params = mod_params.copy()
-            if (
-                "random_state" in current_mod_params
-                and current_mod_params["random_state"] is not None
-            ):
-                current_mod_params["random_state"] += i
+            if current_mod_params.get("random_seed") is not None:
+                current_mod_params["random_seed"] += i
```
dwwest, line 299: Confused about this, why is it adding the model index to the random seed?

openadmet/models/anvil/workflow.py
```diff
@@ -131,15 +133,15 @@ def _train_ensemble(self, X_train_feat, y_train, output_dir, **kwargs):
...
             # Bootstrap data if using bagging, if not specified default False
             if use_bagging:
-                # Set seed for bootstrapping
+                # Use a local generator so bootstrap sampling never mutates global NumPy state
                 logger.info(
-                    f"Using incremented seed={global_seed + i} for bootstrapping"
+                    f"Using incremented seed={bootstrap_seed + i} for bootstrapping"
                 )
-                np.random.seed(global_seed + i)
+                rng = np.random.default_rng(bootstrap_seed + i)
```
dwwest, line 141: Same question here as above?

openadmet/models/features/pairwise.py
```diff
@@ -72,7 +72,7 @@ def downsample_(self, n: int, random_seed: int = 1701):
 
 
 @featurizers.register("PairwiseFeaturizer")
-class PairwiseFeaturizer(FeaturizerBase):
+class PairwiseFeaturizer(DeepLearningFeaturizer):
```
dwwest, line 75: Why was this changed from FeaturizerBase?

dwwest (commented):
Looks good, just a question about the +i in handling bootstrap seeds, should probably be commented so this isn't confusing to future users.

cynthiaxu04 (approved):

openadmet/models/features/pairwise.py
smcolby: DeepLearningFeaturizer already carries the `random_seed` field, FeaturizerBase does not. Switching the base class was the minimal way to gain the field without duplicating it.

PR #576, opened 2026-06-17
Fix ChemProp scheduler configuration, API semantics, and model defaults

smcolby:
## Description

Fixes a series of correctness issues in `ChemPropModel` found during an audit of `configure_optimizers` and the surrounding hyperparameter configuration. All changes are confined to `openadmet/models/architecture/chemprop.py` and `openadmet/models/tests/unit/models/test_chemprop.py`. The unit suite grows from 25 to 30 tests; all 221 unit tests pass.

**Scheduler correctness**

- The Noam `lr_lambda` had an off-by-one at the warmup peak: the old `step < warmup_steps` (strict) caused a tiny discontinuity there because the decay branch handled the peak step instead of the warmup branch. Changed to `step <= warmup_steps` so the warmup branch owns the peak and returns exactly 1.0.
- A `max(1, cooldown_steps)` guard in the decay branch caused one extra step at peak LR when `cooldown_steps == 0`. Removed the guard; the branch is now entered only when `cooldown_steps > 0`.
- When `warmup_epochs=0` and `max_epochs=-1`, both warmup and cooldown steps resolved to zero, causing `lr_lambda` to return `final_factor` (= `final_lr / max_lr`) silently for the entire run. The schedule now holds at `max_lr` in this case and emits a clear warning.
- `ReduceLROnPlateau` was configured with `mode="min"` hardcoded and a single scalar `min_lr=self.final_lr`. Mode now reads from `monitor_metric_mode`. The floor is now computed per param group as `group_lr * (final_lr / max_lr)`, which is correct when `mpnn_lr != max_lr`.
- `steps_per_epoch` was derived from `estimated_stepping_batches`, which is in optimizer-step units (batch count divided by `accumulate_grad_batches`). This underestimates warmup length when gradient accumulation is active. `num_training_batches` (raw batch count) is now the primary source, with `estimated_stepping_batches` as fallback.

**Hyperparameter defaults**

- `aggregation` defaulted to `"norm"` instead of `"mean"`, diverging silently from the published ChemProp baseline.
- `ffn_num_layers` defaulted to 1, reducing the FFN to a linear readout and making `ffn_hidden_dim` irrelevant.
- `reduce_lr_factor` resolved to 0.1 instead of 0.5. Two reductions at 0.1 exhaust most of the LR range; 0.5 is the PyTorch default and more conservative.
- `reduce_lr_patience` resolved to 10 instead of 5, which is coarse for 30-50 epoch runs.

**Hyperparameter bookkeeping**

- The `scheduler` key was not written to `hparams.yaml`, making the logged artifact ambiguous about which schedule was active.
- `mpnn.final_lr` and `mpnn.max_lr` were not assigned in `build()`, so `configure_optimizers` read stale chemprop library defaults rather than user-configured values. For plateau runs this also caused `max_lr` and `final_lr` to appear incorrectly in `hparams.yaml`; those keys are now updated and the irrelevant `init_lr` key is removed.
- `serialize()` omitted resolved LR values, so a config round-tripped through `serialize()` would re-derive different values from defaults. Resolved LRs and structural fields (`depth`, `aggregation`, `ffn_num_layers`, etc.) are now included. `None` entries for inactive-scheduler fields are filtered so they do not appear as explicit nulls.
- Cross-scheduler Pydantic validation could be bypassed by omitting the field from the constructor; the validators now fire on all construction paths.

**Validation and warnings**

- The val-split warning for plateau scheduler fired at `configure_optimizers` time, before Lightning populates `num_val_batches`, causing false positives even when a validation split was configured. Moved to `on_train_start` where the count is reliable. The integer-zero path also now correctly evaluates as no validation rather than being treated as having val.
- `freeze_weights` had an off-by-one when indexing FFN layers.

**Test fixes**

- `test_chemprop_load_weights` called `torch.load` without `weights_only=True`, skipping the deserialization safety check that the production load paths already enforce.
- `test_chemprop_load_weights_invalid_path` passed a string as the `scaler` argument to `build()`, which is silently ignored. The `FileNotFoundError` was raised by `from_foundation` path validation at construction time, not by the argument to `build()`.

**Other**

- Removed a dead import (`build_NoamLike_LRSched`) left over from before `configure_optimizers` was reimplemented.

## Status

### DIFF
--- a/openadmet/models/architecture/chemprop.py
+++ b/openadmet/models/architecture/chemprop.py
@@ -712,12 +712,14 @@ class ChemPropModel(LightningModelBase):
                 mpnn.hparams.pop("warmup_epochs", None)
                 mpnn.hparams.pop("init_lr", None)
 
-            # Pass monitor metric from "model" to "module"
-            # This is necessary to support subclasses of LightningModuleBase, as `monitor_metric`
-            # is needed at the "module" level for use in both `configure_optimizers` and `LightningTrainer`
+            # configure_optimizers and on_train_start below are bound directly onto mpnn
+            # (the LightningModule that Trainer actually calls), not left on self (this
+            # ChemPropModel config wrapper). Inside those bound functions, `self` refers
+            # to mpnn, so every value they read (monitor_metric, the LR/weight-decay
+            # groups, scheduler choice, etc.) must be copied onto the mpnn instance here;
+            # leaving them only on the ChemPropModel would make them unreachable once
+            # the functions run
             mpnn.monitor_metric = self.monitor_metric
-
-            # Attach custom optimization parameters to the MPNN instance
             mpnn.mpnn_weight_decay = self.mpnn_weight_decay
             mpnn.ffn_weight_decay = self.ffn_weight_decay
             mpnn.mpnn_lr = self.mpnn_lr
--- a/openadmet/models/architecture/chemprop.py
+++ b/openadmet/models/architecture/chemprop.py
@@ -231,42 +231,44 @@ class ChemPropModel(LightningModelBase):
         List of metrics to use for evaluation. Default is ["mse", "mae", "rmse"].
     scheduler : str
         Learning rate scheduler ("noam" or "plateau"). Default is "noam".
-    warmup_epochs : int, optional
-        Number of linear-ramp epochs before geometric decay for the Noam scheduler. If None
-        (default), resolves to 2 (matching the original ChemProp paper default). Setting this
-        field with scheduler="plateau" raises ValueError. The Noam schedule shape depends on
-        max_epochs being set correctly in the Lightning Trainer; leaving it at the Lightning
-        default (1000) when actual training runs shorter will under-decay the LR.
-    init_lr : float, optional
-        Starting LR when each param group's base LR equals max_lr. Defaults to max_lr * 0.1.
-        When mpnn_lr or ffn_lr differ from max_lr, the absolute starting LR for that group is
-        group_lr * (init_lr / max_lr), not init_lr; the schedule shape is preserved
-        proportionally around each group's peak.
     max_lr : float
         Peak learning rate (global reference). Default is 1e-3.
     final_lr : float, optional
-        Floor LR when each param group's base LR equals max_lr. Defaults to max_lr * 0.01.
-        Same proportional scaling applies as for init_lr when per-component LRs are set.
+        Floor LR for each param group. Defaults to max_lr * 0.01. When mpnn_lr or ffn_lr
+        differ from max_lr, the absolute floor for that group is group_lr * (final_lr /
+        max_lr); the ratio is preserved proportionally.
     weight_decay : float
         Global weight decay. Default is 0.0.
     mpnn_lr : float, optional
-        Peak learning rate for MPNN. If None, defaults to max_lr.
+        Peak learning rate for the MPNN param group. If None, defaults to max_lr.
     ffn_lr : float, optional
-        Peak learning rate for FFN. If None, defaults to max_lr.
+        Peak learning rate for the FFN param group. If None, defaults to max_lr.
     mpnn_weight_decay : float, optional
-        Weight decay for MPNN. If None, defaults to weight_decay.
+        Weight decay for the MPNN param group. If None, defaults to weight_decay.
     ffn_weight_decay : float, optional
-        Weight decay for FFN. If None, defaults to weight_decay.
+        Weight decay for the FFN param group. If None, defaults to weight_decay.
+    warmup_epochs : int, optional
+        [Noam only] Number of linear-ramp epochs before geometric decay. If None (default),
+        resolves to 2 (matching the original ChemProp paper default). Setting this field
+        with scheduler="plateau" raises ValueError. The schedule shape depends on max_epochs
+        being set correctly in the Lightning Trainer; leaving it at the Lightning default
+        (1000) when actual training runs shorter will under-decay the LR.
+    init_lr : float, optional
+        [Noam only] Starting LR at the beginning of the warmup ramp. Defaults to
+        max_lr * 0.1. When mpnn_lr or ffn_lr differ from max_lr, the absolute starting LR
+        for that group is group_lr * (init_lr / max_lr); the schedule shape is preserved
+        proportionally around each group's peak.
     reduce_lr_factor : float, optional
-        Multiplicative factor applied when plateau is detected (Plateau scheduler only).
-        If None (default), resolves to 0.5. Must be < 1.0. Setting with scheduler="noam" raises ValueError.
+        [Plateau only] Multiplicative factor applied when a plateau is detected. If None
+        (default), resolves to 0.5. Must be < 1.0. Setting with scheduler="noam" raises
+        ValueError.
     reduce_lr_patience : int, optional
-        Epochs with no improvement before LR is reduced (Plateau scheduler only).
-        If None (default), resolves to 5. Setting with scheduler="noam" raises ValueError.
+        [Plateau only] Epochs with no improvement before LR is reduced. If None (default),
+        resolves to 5. Setting with scheduler="noam" raises ValueError.
     monitor_metric_mode : str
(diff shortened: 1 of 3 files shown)


hmacdope (changes requested):
Blocking here as I have a few comments that I need to add, others feel free to review in meantime

openadmet/models/architecture/chemprop.py
```diff
@@ -592,9 +768,47 @@ def train(self, dataloader, scaler=None):
...
+    # set them explicitly. Covers two categories:
+    #   - Structural fields: determine model graph shape and checkpoint compatibility;
+    #     omitting any of these from the artifact makes reloading fragile when the class
+    #     default ever changes
+    #   - Resolved LR fields: computed from max_lr at init time and needed for exact
+    #     schedule reproduction on reload
+    # Scheduler-specific fields (warmup_epochs, reduce_lr_factor, reduce_lr_patience) are
+    # in this set but are None for the inactive scheduler; serialize() drops None entries
+    # so only the active scheduler's fields appear in the artifact
+    _RESOLVED_FIELDS: ClassVar[frozenset[str]] = frozenset(
```
hmacdope: This has a bit of code smell to me, I wonder if `pydantic` has a way to do this more officially

hmacdope (changes requested):

openadmet/models/architecture/chemprop.py
smcolby: [9e000ce](https://github.com/OpenADMET/openadmet-models/pull/576/commits/9e000ce5527d83ac39c286bb42c770995eb35c2e) resolves

cynthiaxu04 (approved):
LGTM

openadmet/models/architecture/chemprop.py
```diff
@@ -9,16 +9,54 @@
...
+        The Lightning trainer the module is attached to.
+
+    Returns
+    -------
+    int
+        Batches per epoch, or a fallback of 1000 with a warning if the
+        trainer cannot report a usable value.
+
+    """
+    steps_per_epoch = getattr(trainer, "num_training_batches", None)
```
hmacdope: `getattr` needed vs just accessing member?

openadmet/models/architecture/chemprop.py
```diff
@@ -71,42 +117,57 @@ def configure_optimizers(self):
...
+                # No warmup and no epoch budget means no way to calibrate decay; hold at max_lr
+                logger.warning(
+                    "noam scheduler with max_epochs=-1 and warmup_epochs=0 cannot calibrate "
+                    "decay; LR will be constant at max_lr for the entire run. "
+                    "Set max_epochs or warmup_epochs > 0 to enable a meaningful schedule."
+                )
+                cooldown_steps = 0
+            else:
+                logger.warning(
+                    "Setting cooldown steps to 100 times the warmup steps for infinite training."
```
hmacdope, line 136: Infinite training?

openadmet/models/architecture/chemprop.py
```diff
@@ -116,6 +177,34 @@ def lr_lambda(step: int):
...
+def _warn_if_no_val_dataloader(self) -> None:
+    """
+    Emit a warning when plateau scheduler has no validation dataloader.
+
+    Bound to on_train_start so num_val_batches is fully populated by Lightning
+    before this check runs. configure_optimizers fires too early and sees an
+    empty list even when a val split is configured.
+    """
+    if self.scheduler != "plateau":
+        return
```
hmacdope, line 190: I think the application of this only to `plateau` scheduler needs clarification in the name of the function.

openadmet/models/architecture/chemprop.py
```diff
@@ -116,6 +177,34 @@ def lr_lambda(step: int):
...
+    num_val_batches = getattr(self.trainer, "num_val_batches", None)
+    if num_val_batches is None:
+        # Trainer did not expose the attribute; assume val exists
+        return
+    if hasattr(num_val_batches, "__iter__"):
+        has_val = any(n > 0 for n in num_val_batches)
```
hmacdope: This is a bit opaque to me

openadmet/models/architecture/chemprop.py
```diff
@@ -135,97 +224,143 @@ class ChemPropModel(LightningModelBase):
...
-    depth: int = 3
-    message_hidden_dim: int = 300
-    ffn_hidden_dim: int = 300
-    ffn_num_layers: int = 1
-    normalized_targets: bool = True
-    batch_norm: bool = False
-    dropout: float = 0.0
+    # Structural fields are tagged resolved=True: they determine model graph shape
+    # and checkpoint compatibility, so serialize() always persists them regardless
+    # of whether the user set them explicitly (see _resolved_fields below)
```
hmacdope, line 327: Is it possible to do this with a `partial` or something like that to avoid having the field everywhere? all good if not

openadmet/models/architecture/chemprop.py
```diff
@@ -292,28 +430,43 @@ def resolve_hyperparameters(self) -> "ChemPropModel":
...
+        # Fill scheduler-specific defaults only for the active scheduler
+        if self.scheduler == "noam":
+            if self.warmup_epochs is None:
+                self.warmup_epochs = 2
+        elif self.scheduler == "plateau":
+            if self.reduce_lr_factor is None:
+                self.reduce_lr_factor = 0.5
+            if self.reduce_lr_patience is None:
+                self.reduce_lr_patience = 5
+
```
hmacdope: Hardcode? should this be configurable?

hmacdope (changes requested):
Sorry this took so long, few queries, but mostly looks good. Would be good to also show some testing

openadmet/models/architecture/chemprop.py
smcolby: `max_epochs=-1` is Lightning's actual sentinel for "no epoch cap" (trainer runs until max_steps, early stopping, or manual interruption). So this guards `cooldown_epochs = max_epochs - warmup_epochs` when `max_epochs=-1`, i.e. nothing stops a user from picking `scheduler="noam"` with `max_epochs=-1`.

openadmet/models/architecture/chemprop.py
smcolby: Yeah this was cryptic... `num_val_batches` is either a count (one validation dataloader, or none) or a list of counts (one per dataloader, when there are several), so the code has to check which shape it got before deciding whether any validation data exists. Added comments and also a more intuitive check on the if statement.

openadmet/models/architecture/chemprop.py
smcolby: Yes, implemented.

openadmet/models/architecture/chemprop.py
smcolby: These defaults live in a validator instead of the field declaration because the field also needs to detect "the user typed this value for the wrong scheduler" as an error, and that's only possible if unset fields stay None until resolved here, rather than never being distinguishable from user input.

So these are configurable, and they fallback to the (hardcoded) defaults here.

openadmet/models/architecture/chemprop.py
smcolby: Renamed to `_warn_if_plateau_missing_val_dataloader`

openadmet/models/architecture/chemprop.py
smcolby: Docstrings also reflect this:
- warmup_epochs (chemprop.py:296-301): "If None (default), resolves to 2 (matching the original ChemProp paper default)."
- reduce_lr_factor (chemprop.py:307-310): "If None (default), resolves to 0.5."
- reduce_lr_patience (chemprop.py:311-313): "If None (default), resolves to 5."

openadmet/models/architecture/chemprop.py
smcolby: Swapped for `steps_per_epoch = trainer.num_training_batches`. The only reason this was `getattr` before was due to compliance with a test that didn't use a full trainer instance. That test has been tightened up, so we can be more explicit here as a result.

hmacdope (approved):

PR #577, opened 2026-06-18
Fix UnscaleTransform identity when no scaler is provided

smcolby:
## Description

`_get_output_transform` returns an `UnscaleTransform` placeholder when `normalized_targets=True` and no scaler is supplied, with the expectation that the transform's buffers will be overwritten when checkpoint weights are loaded. It was constructed as `UnscaleTransform([1]*n_tasks, [0]*n_tasks)`, which computes `X * 0 + 1 = 1.0` at inference regardless of what the model outputs. The correct identity is mean=0, scale=1, giving `X * 1 + 0 = X`. Changed to `UnscaleTransform([0]*n_tasks, [1]*n_tasks)`.

The existing test for this path only asserted `isinstance`; it is extended to verify in eval mode that a non-trivial input passes through unchanged.

## Status

### DIFF
--- a/openadmet/models/architecture/chemprop.py
+++ b/openadmet/models/architecture/chemprop.py
@@ -412,9 +412,9 @@ class ChemPropModel(LightningModelBase):
         if scaler is not None:
             output_transform = nn.UnscaleTransform.from_standard_scaler(scaler)
         elif self.normalized_targets:
-            # Expects the targets to be normalized, likely to be loaded from state dict
+            # Identity placeholder (mean=0, scale=1); real parameters are loaded from checkpoint
             output_transform = nn.UnscaleTransform(
-                [1] * self.n_tasks, [0] * self.n_tasks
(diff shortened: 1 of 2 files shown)


hmacdope (approved):
LGTM

hmacdope:
@smcolby can we just confirm no changes here with a quick test otherwise good to go

smcolby:
> @smcolby can we just confirm no changes here with a quick test otherwise good to go

But there _are_ changes, I just don't think this path ever actually triggers since we use scaling everywhere. The changes are necessary, though, as before it was using `mean=1, scale=0` instead of `mean=0, scale=1`. 

So if we somehow engineer a test of "old way", it'll produce the wrong result, thus appearing as if "new way" is a regression. 

Pattern is (directly copied from chemprop docs): `chemprop.nn.transforms.UnscaleTransform(mean, scale, pad=0)`. We had `UnscaleTransform(1, 0)` previously, now correctly `UnscaleTransform(0, 1)`.

PR #592, opened 2026-08-17
Add PCA dimensionality reduction to the anvil pipeline

smcolby:
## Description

Adds PCA dimensionality reduction as a first-class step in the anvil pipeline. `PCATransform` occupies a new `transform` slot in the procedure section, between featurization and training. With an integer `n_components` it runs a single PCA over the full feature matrix; with a dict it reduces each featurizer block to its own component count (e.g. fingerprints to 256 dims and descriptors to 32), resolving the block layout from the featurizer via a new `feature_blocks` protocol. A per-block count of `null` passes that block through unreduced, for blocks whose columns carry meaning one by one and would only be rotated by a PCA; every block still needs an entry, so passthrough is explicit and a mistyped key is still caught. An optional `impute_strategy` handles NaNs ahead of the reduction since PCA cannot see NaNs itself.

Fitting happens on the train partition only, the transform is applied to every partition, and the fitted transform is persisted beside the model (schema-versioned `transform.pickle`) so that decoupled inference sees exactly the features training saw: `predict()` loads the artifact and applies it, and raises loudly instead of silently skipping when a model trained with a transform has no artifact. The `transform` section accepts a single transform or an ordered list, so column operations such as median imputation followed by per-block PCA can be stacked. Transforms are consumed by tabular-input workflows only: the sklearn workflow rejects featurizers that emit DataLoaders, and the lightning workflow rejects transforms, so ChemProp pairings fail at construction with a clear message.

**The transform and its block protocol**

- New `PCATransform` in the transforms registry: int or per-block `n_components` (an int, or `null` for passthrough), optional mean/median imputation, per-entry `random_seed`; validates block keys against the featurizer layout, exact width coverage of the raw matrix, and component counts against the fitting data, where a count may go up to the block rank as sklearn itself allows
- `fit_transforms` / `transform_features` helpers to fit and apply a single transform or an ordered sequence, threading each output into the next
- `FeaturizerBase.feature_blocks` reporting the raw featurizer's column layout keyed by registry name; `FeatureConcatenator` flattens nested concatenations in emit order

**Training and inference wiring**

- Anvil workflow: probes the block layout on train rows, fits the transform sequence on train only, applies it to train/val/test/full, and saves the fitted transforms beside the model
- Anvil specification and YAML parsing: `transform` accepts a `TransformSpec` or a list of them; the global seed fills entries that set none
- Inference: `load_anvil_model_and_metadata` returns the fitted transform as a fifth element (existing in-package callers updated); `predict()` applies it after featurization, normalizes 1D single-row output, and checks that row count is preserved
- `FeatureConcatenator` accepts the list entry form used in recipe YAML and rejects duplicate featurizer classes, whose identical block keys could not be disambiguated
- Post-hoc comparison label builder handles the list feature section form
- `ImputeTransform.fit` returns `self` as its docstring always claimed (chained usage was previously broken)

**Tests and docs**

- Unit tests for PCATransform, the sequence helpers, block layouts, workflow/specification wiring, and inference; two new CPU integration recipes (single PCA over ECFP, and stacked imputation plus per-block PCA over a fingerprint plus 2D-descriptor concatenation)
- Docs: API pages for the transforms module and a `transform` section (single and per-block examples) in the anvil reference

**Notes for reviewers**

- Model directories trained with a transform by an early dev build of this feature lack `transform.pickle` and will now hard-error at inference until retrained (previously inference would have consumed the wrong features silently).
- Per-block PCA assumes a stable column layout between train and inference batches; featurizers whose width is batch-dependent (mordred in some environments) trip the fit-time width guard, by design and documented in the docstring.

## Status

---


docs/anvil_reference.rst
```diff
@@ -244,35 +253,140 @@ while traditional machine learning models return a a 2D ``NumPy`` array or ``pan
...
 
 .. code-block:: yaml
+
   feat:
     type: ChemPropFeaturizer
     params: {}
 
+Transform
+~~~~~~~~~
+The optional ``transform`` section transforms the featurized feature matrices after
```
dwwest: Featurized feature matricies

openadmet/models/anvil/specification.py
```diff
@@ -210,12 +211,7 @@ def read_split(resource: str, split_name: str) -> pd.DataFrame:
             [read_split(resource, split) for resource, split in splits_to_read]
         )
 
-        target_cols = (
```
dwwest, line 213: Does this still work as expected? It looks like passing a single string will cause an error without this line?

openadmet/models/anvil/specification.py
```diff
@@ -700,7 +691,15 @@ class ProcedureSpec(SpecBase):
     model: ModelSpec
     ensemble: EnsembleSpec | None = None
     train: TrainerSpec
-    transform: Optional[TransformSpec] = None  # Optional transform step
+    transform: list[TransformSpec] | None = None  # Optional transform sequence
+
+    @field_validator("transform", mode="before")
+    @classmethod
+    def _wrap_single_transform(cls, value):
```
dwwest, line 699: This function is repeated several times, is it possible to have it only be written in one file and imported?

openadmet/models/features/combine.py
```diff
@@ -21,48 +41,197 @@ class FeatureConcatenator(FeaturizerBase):
...
+            # to check against the declared list[FeaturizerBase]
             return value
 
-        # Sort the featurizers by class name
-        return sorted(processed_featurizers, key=lambda f: f.__class__.__name__)
+        return processed_featurizers
+
+    @field_validator("featurizers", mode="after")
+    @classmethod
+    def reject_duplicates_and_sort(cls, value):
```
dwwest, line 148: I'm not sure I understand the sorting here... how did the transform/blocks cause this to be necessary?

openadmet/models/transforms/pca.py
```diff
@@ -0,0 +1,347 @@
...
+    transforms,
+)
+
+
+@transforms.register("PCATransform")
+class PCATransform(TransformBase):
+    """
+    Apply principal component analysis to one or more feature blocks.
+
+    With an int ``n_components``, a single PCA is fitted over the entire
```
dwwest: I think a deeper explanation for the reason for splitting into blocks would help here.

dwwest (commented):
I need to chat about the blocks again I think, I'm not sure I totally understand that.

docs/anvil_reference.rst
smcolby: Will fix

openadmet/models/anvil/specification.py
smcolby: Will verify that this is being handled correctly, likely pydantic?
```python
"""
target_cols : list[str]
        The target column(s) in the dataset. A bare string is accepted and
        wrapped into a one-element list.
"""
```

openadmet/models/anvil/specification.py
smcolby: Will investigate

openadmet/models/features/combine.py
smcolby: I will improve the docstring comment

openadmet/models/transforms/pca.py
smcolby: I will improve the docstring

docs/anvil_reference.rst
```diff
@@ -244,35 +253,140 @@ while traditional machine learning models return a a 2D ``NumPy`` array or ``pan
...
+      type: FingerprintFeaturizer
+      params:
+        fp_type: "ecfp:4"
+    transform:
+      type: PCATransform
+      params:
+        n_components: 256
+        random_seed: 42
+
+Per-block PCA: give ``n_components`` as a mapping from featurizer name to component count. Each featurizer in the
```
dwwest, line 331: I think here is a good place to have a specific example written out to make this more clear.

openadmet/models/anvil/workflow.py
```diff
@@ -309,24 +348,42 @@ def run(
...
             if X_test is not None:
-                X_test_feat = self.transform.transform(X_test_feat)
+                X_test_feat = transform_features(self.transform, X_test_feat)
                 zarr.save(data_dir / "X_test_feat_transformed.zarr", X_test_feat)
 
             # Whole dataset
-            X_feat = self.transform.transform(X_feat)
+            X_feat = transform_features(self.transform, X_feat)
+
+            # Persist the fitted transforms next to the model so decoupled
```
dwwest: This is one of those comments that's hard to read. More specifics here.

openadmet/models/anvil/workflow_base.py
```diff
@@ -73,6 +83,50 @@ class AnvilWorkflowBase(BaseModel):
     debug: bool = False
     resolved_output_dir: Path | None = None
 
+    @model_validator(mode="after")
+    def check_transform_block_keys(self):
+        """
+        Check per-block transforms against the featurizer's block layout.
+
+        A transform configured per block (e.g. PCATransform with a dict
```
dwwest: Here, too, a specific example would help.

openadmet/models/transforms/pca.py
```diff
@@ -0,0 +1,347 @@
...
+        # Imputation runs inside the pipeline so it is fitted on train rows only,
+        # the same as the PCA that follows it
+        if self.impute_strategy != "none":
+            steps.append(("impute", SimpleImputer(strategy=self.impute_strategy)))
+
+        steps.append(("pca", PCA(n_components=dims, random_state=self.random_seed)))
+
+        return Pipeline(steps)
+
+    def _check_2d(self, X: np.ndarray) -> None:
```
dwwest, line 139: Why is this check necessary?

openadmet/models/transforms/pca.py
```diff
@@ -0,0 +1,347 @@
...
+        # Per-block form: walk the blocks in column order, fitting one PCA each
+        if isinstance(self.n_components, dict):
+            blocks = self._validate_blocks(feature_blocks, n_features)
+            fitted_blocks = []
+            cursor = 0
+
+            for key, width in blocks:
+                dims = self.n_components[key]
+
+                # A passthrough block is recorded with no pipeline, so it keeps
```
dwwest: I think based on our discussion yesterday, I agree that the long in-line comments actually make the code harder to read.

dwwest (approved):
Looks good, I think mostly comments to make things clearer. Example use-cases would really help

docs/anvil_reference.rst
smcolby: The example comes right after this paragraph now.

openadmet/models/anvil/specification.py
smcolby: This works, by way of `DataSpec.check_target_cols_input` validator

openadmet/models/anvil/specification.py
smcolby: This has been streamlined.

smcolby: Extracted to a new openadmet/models/anvil/utils.py

openadmet/models/anvil/workflow.py
smcolby: Shortened to:
```python
# Inference needs the fitted state, which the recipe YAML does not have implicitly
```

openadmet/models/anvil/workflow_base.py
smcolby: The example lives in the docs, but I added the 3 cases in prose here.

openadmet/models/features/combine.py
smcolby: The sorting was already present, just built on here. I made the text more clear:
>Blocks are sorted by class name so that featurization and transformation stay consistent across recipes naming the same featurizers. Same-class featurizers are rejected because per-block transforms address blocks by that name and could not tell two apart.

openadmet/models/transforms/pca.py
smcolby: I lost the origin of this comment, sorry...

openadmet/models/transforms/pca.py
smcolby: These have been shortened

openadmet/models/transforms/pca.py
```diff
@@ -0,0 +1,351 @@
...
+        of None leaves that block unreduced.
+    impute_strategy : str
+        Optional imputation applied to each block before its PCA. Options are
+        'none' (default), 'mean', or 'median'.
+    random_seed : int or None
+        Random seed for the PCA solvers, by default 42, matching the
+        procedure-level default a workflow would otherwise fill in. Threaded to
+        ``random_state``, which matters because sklearn picks the randomized
+        solver for wide matrices with a large reduction, exactly the per-block
+        fingerprint case. Set to None only to opt into an unseeded solver.
```
smcolby: This is way too long to describe `random_seed`

PR #598, opened 2026-08-21
Add TrainedModelFeaturizer for featurizing with a pretrained model

smcolby:
## Description

Adds `TrainedModelFeaturizer`, which loads an already-trained Anvil model from
disk and emits its predictions as feature columns for a downstream model. The
loaded model is run in inference mode and never refitted, so it acts as a frozen
feature source.

The motivating case is transferring signal from an abundant low-fidelity endpoint
to a scarce high-fidelity one: a model trained on primary-screen log2 fold change
supplies its predictions as features to a model trained on dose-response pEC50.
The target pipeline is PCA(CheMeleon embedding) + log2FC predictions -> TabICL.

```yaml
feat:
  type: TrainedModelFeaturizer
  params:
    model_dir: /path/to/pxr_log2fc_model
```

Design points a reviewer should weigh:

- The pretrained model brings its own featurizer, so this featurizer takes SMILES
  rather than features, and reports the row indices that featurizer kept. A
  `FeatureConcatenator` can then intersect those indices with other blocks as it
  does for any featurizer that drops rows.
- It also reproduces the rest of the pretrained model's inference path. Since
  #592, `load_anvil_model_and_metadata` returns the model's fitted transform, and
  a model trained in PCA space has to be predicted on in PCA space. The transform
  is cached alongside the model and applied to the featurizer's output before
  predicting.
- `return_std` appends the ensemble standard deviation as a second block of
  columns, off by default. Only an ensemble has one, and whether the model is an
  ensemble is answered by `recipe_components/procedure.yaml`, so that pairing is
  checked when the recipe is parsed rather than after a full featurization pass.
  `model_dir` is validated the same way: at construction, not at first use.
- Columns are one per task, with the standard deviation block after the
  prediction block (`pred_task0, pred_task1, std_task0, std_task1`), so a
  single-task model contributes one column by default.
- The model is deserialized once and cached on the instance, so featurizing
  several partitions does not reload it.
- The block key is the class name, matching the existing featurizer convention.
  A known consequence is that two `TrainedModelFeaturizer` entries cannot coexist
  in one concatenator, given the duplicate-key check. Left as is deliberately, in
  keeping with the convention.

**How to verify**

- Unit tests: `pytest openadmet/models/tests/unit -q`. On this branch that runs
  360 passed, 5 skipped locally. Integration tests were not run locally and are
  left to CI.
- New coverage is in `openadmet/models/tests/unit/features/test_trained_model.py`
  (11 tests): prediction emission for a single-task model, ensemble mean, the
  prediction-then-standard-deviation column ordering, index propagation when the
  pretrained featurizer drops an unparseable SMILES, application of the
  pretrained model's fitted transform, construction-time rejection of `return_std`
  against a non-ensemble model, a bad `model_dir`, a recipe with no
  `procedure.yaml`, registry round-trip, composition inside a
  `FeatureConcatenator`, and load-once caching.
- The transform test is the one worth a close look. It builds a model trained in
  PCA space (`transform_model_dir` in `conftest.py`) and checks the featurizer's
  output against predictions reached independently through `transform_features`.
  Stubbing out the transform branch makes it fail, so it does catch the bug it
  was written for.
- `ruff check openadmet/models/` and `ruff format --check` are clean.
- Docs: new `docs/_api/api/featurization/trained_model.rst` in the toctree, plus a
  table row, prose, and a YAML example in `docs/anvil_reference.rst`.

**Notes for reviewers**

This branch started as one of four independent branches off `main`. The other
three have since merged, so it now carries a merge of `main` and the integration
work that merge exposed: the transform threading described above, plus a move to
the list form for `FeatureConcatenator` after #592 deprecated the whole-field
dict form.

Selecting a subset of a multitask model's output columns is deliberately not
supported. Someone who wants one CYP as a feature can point at a model trained on
that CYP.

Guarding against data leakage between the featurizing model and the end-to-end
production model is left to the user; this PR adds no check or warning for it by
design.

## Status


openadmet/models/features/trained_model.py
```diff
@@ -0,0 +1,256 @@
...
+            raise ValueError(
+                f"Model directory {value} has no recipe_components directory, so it "
+                "is not a trained Anvil model."
+            )
+
+        return value
+
+    @field_validator("outputs")
+    @classmethod
+    def reject_duplicate_outputs(cls, value: list[str]) -> list[str]:
```
smcolby: Why would this ever hit?

openadmet/models/features/trained_model.py
```diff
@@ -0,0 +1,256 @@
...
+        if "std" in self.outputs:
+            mean, std = model.predict(
+                X_feat, accelerator=self.accelerator, return_std=True
+            )
+        else:
+            mean = model.predict(X_feat, accelerator=self.accelerator)
+            std = None
+
+        # Lay the requested quantities out in the configured order, each
+        # contributing one column per task
```
smcolby: Check logic here

openadmet/models/features/trained_model.py
```diff
@@ -0,0 +1,256 @@
...
+            return self
+
+        # Reading the recipe is a few kilobytes of YAML; deserializing the model
+        # to ask the same question would cost orders of magnitude more
+        procedure_path = self.model_dir / "recipe_components" / "procedure.yaml"
+        if not procedure_path.is_file():
+            raise ValueError(
+                f"Model directory {self.model_dir} has no recipe_components/"
+                "procedure.yaml, so it is not a trained Anvil model."
+            )
```
smcolby: L153 - 160 check that it's a trained anvil model, which is beyond the scope of checking for `std`. We should validate the anvil model no matter what, and detect if it supports `std` if `std` is requested. Either generalize the name of this validator, or have separate validators.

Oh, looking above, there already *is* a `validate_model_dir`, so the check for anvil model is redundant here?

openadmet/models/features/trained_model.py
smcolby: Specifically, the following is a bit opaque, especially with the static method `_as_columns`:

```python
blocks = {"mean": mean, "std": std}
columns = [self._as_columns(blocks[name]) for name in self.outputs]

return np.concatenate(columns, axis=1).astype(np.float64), np.asarray(indices)
```

openadmet/models/features/trained_model.py
```diff
@@ -0,0 +1,256 @@
...
+        """
+        model, feat, _, _ = self._load()
+
+        # The pretrained model owns its featurization, so it consumes SMILES and
+        # reports which of them it managed to featurize
+        feat_data = feat.featurize(smiles)
+        X_feat, indices = feat_data[0], feat_data[1]
+
+        # A model reports its spread only on request, and only an ensemble has
+        # one; the validator has already established that pairing is possible
```
smcolby: Comment too long: "Report std if requested" is better

openadmet/models/features/trained_model.py
```diff
@@ -0,0 +1,256 @@
...
+        tuple
+            Tuple of (features, indices). Features has shape
+            (n_featurized, len(outputs) * n_tasks); indices are the positions
+            in the input that the pretrained model's featurizer kept.
+
+        """
+        model, feat, _, _ = self._load()
+
+        # The pretrained model owns its featurization, so it consumes SMILES and
+        # reports which of them it managed to featurize
```
smcolby: Comment too long: "Featurize using pretrained model's featurizer" is better

openadmet/models/features/trained_model.py
```diff
@@ -0,0 +1,256 @@
...
+                load_anvil_model_and_metadata,
+            )
+
+            model, feat, metadata, data_spec = load_anvil_model_and_metadata(
+                self.model_dir
+            )
+
+            # Inference never shuffles, but pin it off in case the pretrained
+            # model's featurizer would otherwise reorder rows out of step with
+            # the indices it reports
```
smcolby: Is this over-guarded? If it must stay, shorten comment to "Ensure featurizer doesn't shuffle"

openadmet/models/features/trained_model.py
```diff
@@ -0,0 +1,256 @@
...
+
+        Returns
+        -------
+        tuple
+            The loaded (model, featurizer, metadata, data spec).
+
+        """
+        if self._loaded is None:
+            # Imported here so the featurizer registry does not pull in the
+            # inference module, and its dependencies, at import time
```
smcolby: Comment too long: "Defer import" is better

openadmet/models/features/trained_model.py
```diff
@@ -0,0 +1,256 @@
...
+        if procedure.get("ensemble") is None:
+            raise ValueError(
+                f"outputs includes 'std', but the model at {self.model_dir} is not an "
+                "ensemble and has no spread to report. Use outputs: [mean], or point "
+                "at an ensemble model."
+            )
+
+        return self
+
+    def _load(self) -> tuple:
```
smcolby: Why is this private? Convention elsewhere in repo is `load`.

docs/anvil_reference.rst
```diff
@@ -265,6 +267,28 @@ You can also combine multiple traditional ML featurizers using the ``FeatureConc
...
+       outputs: [mean]
+
+The pretrained model brings its own featurizer, so the ``TrainedModelFeaturizer`` takes SMILES rather than features.
+Molecules that the pretrained featurizer cannot parse are reported through the returned index array, the same as any
+other featurizer.
+
+``outputs`` selects which per-task quantities become columns, in column order. ``mean`` is the model's prediction and
+is the default. ``std`` is the spread across ensemble members, so it requires the referenced model to be an ensemble;
+requesting it from a single model raises when the recipe is parsed. The emitted width is ``len(outputs)`` columns per
+target column of the pretrained model, laid out output-major and task-minor.
```
smcolby: This sort of makes it sound like a `TrainedModelFeaturizer` *must* support `std`, not that it is optional and only for ensembles.

openadmet/models/features/trained_model.py
```diff
@@ -0,0 +1,256 @@
...
+
+        # Lay the requested quantities out in the configured order, each
+        # contributing one column per task
+        blocks = {"mean": mean, "std": std}
+        columns = [self._as_columns(blocks[name]) for name in self.outputs]
+
+        return np.concatenate(columns, axis=1).astype(np.float64), np.asarray(indices)
+
+    @staticmethod
+    def _as_columns(values: np.ndarray) -> np.ndarray:
```
smcolby: Is this not redundant with other shape-management utilities elsewhere in the repo?

openadmet/models/features/trained_model.py
smcolby: I also don't know what the accompanying comment means.

openadmet/models/features/trained_model.py
```diff
@@ -0,0 +1,256 @@
...
+
+    model_dir: Path = Field(
+        ..., description="Directory of the trained model to featurize with"
+    )
+    outputs: list[Literal["mean", "std"]] = Field(
+        default=["mean"],
+        min_length=1,
+        description="Per-task quantities to emit as feature columns, in column order",
+    )
+    accelerator: str = "cpu"
```
smcolby: Shouldn't this default to "auto"?

openadmet/models/features/trained_model.py
```diff
@@ -0,0 +1,230 @@
...
+        'auto', which uses a GPU where one is available and falls back to CPU.
+
+    """
+
+    type: ClassVar[str] = "TrainedModelFeaturizer"
+
+    model_dir: Path = Field(
+        ..., description="Directory of the trained model to featurize with"
+    )
+    outputs: list[Literal["mean", "std"]] = Field(
```
dwwest: Can this be given a more descriptive name?

openadmet/models/features/trained_model.py
```diff
@@ -0,0 +1,230 @@
...
+        duplicates = sorted({name for name in value if value.count(name) > 1})
+        if duplicates:
+            raise ValueError(f"Duplicate outputs: {duplicates}.")
+
+        return value
+
+    @model_validator(mode="after")
+    def check_std_is_available(self):
+        """
+        Check the pretrained model can produce a spread when 'std' is requested.
```
dwwest: Why would we ever need this for pretrained models? Would this hit either?

openadmet/models/features/trained_model.py
```diff
@@ -0,0 +1,230 @@
...
+        tuple
+            Tuple of (features, indices). Features has shape
+            (n_featurized, len(outputs) * n_tasks); indices are the positions
+            in the input that the pretrained model's featurizer kept.
+
+        """
+        model, feat = self._load_pretrained_model()
+
+        # Featurize using pretrained model's featurizer
+        X_feat, indices = feat.featurize(smiles)[:2]
```
dwwest: What is [:2] doing?

openadmet/models/tests/unit/features/test_trained_model.py
```diff
@@ -0,0 +1,191 @@
...
+
+    The NullFeaturizer used by the shared fixtures keeps every row, so it cannot
+    exercise index propagation. This model predicts 5.0 for any input
+    (tag=FP, target=task_0).
+    """
+    model_dir = tmp_path_factory.mktemp("fingerprint_model")
+    recipe_dir = model_dir / "recipe_components"
+    recipe_dir.mkdir(parents=True)
+
+    with open(recipe_dir / "metadata.yaml", "w") as f:
```
dwwest: Why are we doing direct opening of files within the main test file?

dwwest (approved):
Looks good, couple clarifications

openadmet/models/features/trained_model.py
smcolby: Do you have any suggestions?

openadmet/models/features/trained_model.py
smcolby: If it's an ensemble, the `predict` method can produce a standard deviation, which could ostensibly be passed as a feature here. Very edge case, but it's there if we want it.

openadmet/models/features/trained_model.py
smcolby: Good question. Had to ask our mutual friend. Featurizers return either a 2-tuple (molfeat_fingerprint.py:79`, molfeat_properties.py:88` return `feat, indices`) or a 4-tuple (`chemprop.py:193` returns `dataloader, indices, scaler, dataset`; `pairwise.py:198` similar). `[:2]` takes the first two, the ones needed.

More clearly it would look like:
```python
# Featurize using pretrained model's featurizer
feat_data = feat.featurize(smiles)

# Featurizers return (features, indices) or (dataloader, indices, scaler, dataset)
X_feat, indices = feat_data[0], feat_data[1]
```

openadmet/models/tests/unit/features/test_trained_model.py
smcolby: Good catch. Working on this.

openadmet/models/features/trained_model.py
smcolby: You flagging this did catch a bug in the docstrings elsewhere though. It was saying a `std` request for a model that can't produce it would return an all-NaN column, which isn't true. Now fixed!

openadmet/models/tests/unit/features/test_trained_model.py
smcolby: Other `model-dir` fixtures already lived in `conftest.py`. Generalized the helper there and moved this one in with them.

openadmet/models/features/trained_model.py
smcolby: How about `model_outputs`?

smcolby: Ended up replacing the field rather than renaming it. `std` never comes without a prediction, so no need to name "outputs" / "predictions" etc. at all. It defaults to `False`, so will assume the typical workflow of no ensemble / no `std`. 

```yaml
params:
  model_dir: /path/to/model
  return_std: true
```

Oh, and it's named `return_std` to match how `CommitteeRegressor`, the workflow, and inference already spell it on `predict`.

PR #599, opened 2026-09-03
Rename butina cutoff kwarg to dist_cutoff

smcolby:
## Description

The `useful_rdkit_utils` dependency is pinned to a moving branch (`git+https://github.com/PatWalters/useful_rdkit_utils.git@master`). Upstream renamed the `get_butina_clusters` parameter from `cutoff` to `dist_cutoff`. Because the pin is a branch rather than a fixed ref, any freshly built environment now resolves the newer upstream commit, and the call site in `openadmet/models/split/cluster.py` raised:

```
TypeError: get_butina_clusters() got an unexpected keyword argument 'cutoff'
```

This surfaced as `test_cluster_split_synthetic_data[butina]` failing on `main` after a clean CI environment build, even though it passed on caches holding the older upstream. This changes only the keyword at the call site to match the current upstream API. The dependency ref is intentionally left unpinned per the scope of this fix; the upstream return type also changed from `List[int]` to `np.ndarray`, which the existing downstream cluster handling accepts unchanged.

## Status

### DIFF
--- a/openadmet/models/split/cluster.py
+++ b/openadmet/models/split/cluster.py
@@ -71,7 +71,7 @@ class ClusterSplitter(SplitterBase):
         if self.method == "butina":
             from useful_rdkit_utils import get_butina_clusters
 
-            clusters = get_butina_clusters(X, cutoff=self.butina_cutoff)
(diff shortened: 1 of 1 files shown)


dwwest (approved):
LGTM

PR #601, opened 2026-09-09
fix: resolve accelerator to device in TabPFN model classes

### LINKED ISSUE #589: Inconsistent accelerator vs device handling across models
## Summary

Public model APIs use `accelerator` with values like `"cpu"`/`"gpu"`/`"auto"`, but the underlying frameworks (PyTorch, Lightning, TabICL, TabPFN) each expect their own native `device` spelling (`"cuda"`, `"mps"`, `"xla"`, or `None`). The mapping between the two is handled inconsistently, sometimes correctly, sometimes not at all, across model classes.

## Impact

- **Confirmed crash**: passing `accelerator="auto"` (the default) to `TabICLRegressorModel`/`TabICLClassifierModel` raised `RuntimeError: Expected one of cpu, cuda, ... device type at start of device string: auto` from `torch.device("auto")`, since `"auto"` was passed straight through as a literal string instead of being resolved. Fixed in `tabicl.py` (see below); the same bug is still present in `tabpfn.py`.
- **Silent gaps**: `"tpu"`/`"mps"` accelerators were unreachable in some models even though the underlying framework supports them, because the public `accelerator` field was a narrow `Literal` that only allowed `"cpu"`/`"gpu"`/`"auto"`.
- **Inconsistent public API within a single file**: `tabpfn.py` currently declares two different `Literal` types for the same concept across its two model classes: one uses `Literal["cpu", "gpu", "auto"]`, the other `Literal["cpu", "cuda", "auto"]`. Callers must pass a different spelling depending on which class they're using.

## Current state

Two of the five files listed in the original issue have been fixed, using a converged pattern rather than a single shared cross-framework helper (see **Design decision** below):

| File | Status | Notes |
|---|---|---|
| `openadmet/models/architecture/chemprop.py` | Fixed | `_resolve_device(accelerator)`: alias dict (`"gpu"→"cuda"`, `"tpu"→"xla"`), `"auto"` resolved eagerly via Lightning's accelerator connector (`chemprop.py` runs under `lightning.pytorch`). |
| `openadmet/models/features/chemeleon_embedding.py` | Fixed | Reuses `chemprop._resolve_device` rather than duplicating the mapping (previously had its own narrower `_normalize_accelerator` that only handled `"gpu"`). |
| `openadmet/models/architecture/tabicl.py` | Fixed | `_resolve_device(accelerator)`: same alias dict, but `"auto"` maps to `None`, not a resolved string, since TabICL's own `device=None` triggers its own `torch.cuda.is_available()` check internally. `accelerator` widened from `Literal["cpu","gpu","auto"]` to `str` so `"mps"`, `"cuda:0"`, etc. are reachable. Previously also carried a hand-maintained whitelist of TabICL constructor kwargs that silently dropped unknown ones; now unsupported kwargs raise `TypeError` from TabICL's own constructor instead. |
| `openadmet/models/architecture/tabpfn.py` | **Not fixed** | Same `"auto"`-passthrough bug as TabICL had (`accelerator = self.accelerator if self.accelerator != "gpu" else "cuda"`, then passed as `device=accelerator`). Two model classes in this file also disagree with each other on the `accelerator` `Literal` values (see Impact above). |
| `openadmet/models/inference/inference.py` | No change needed | Only plumbs the `accelerator` string through to `model.build()`/`model.predict()`; does not construct a `torch.device` itself. The fix belongs at each model class's boundary. |
| `openadmet/models/cli/predict.py` | No change needed | Same as above, just forwards the CLI flag. |

## Design decision: no single shared helper

The original recommendation was one normalization helper used at every accelerator to device boundary. In practice, `"auto"` resolution is framework-specific and a single shared helper would either hide that divergence or bake in one framework's assumptions:

- Lightning-based models (ChemProp) want `"auto"` eagerly resolved to a concrete device string via Lightning's own accelerator connector.
- TabICL wants `"auto"` passed through as `None` so it runs its own `cuda.is_available()` check, unaware of Lightning.
- TabPFN's own auto-detection behavior needs to be confirmed before deciding which of the two patterns above it should follow.

The converged pattern instead is: each model module owns a small local `_ACCELERATOR_ALIASES` dict plus a `_resolve_device(accelerator)` function with the same alias shape, but each free to resolve `"auto"` according to what its underlying framework expects. `chemeleon_embedding.py` reuses `chemprop.py`'s version directly since both run under Lightning; `tabicl.py` has its own since TabICL's semantics differ.

## Remaining work

1. Apply the same fix shape to `tabpfn.py`:
   - Confirm what TabPFN's own `device` parameter expects for auto-detection (`None`, `"auto"`, or something else) before choosing the `"auto"` resolution strategy.
   - Add a local `_resolve_device` (or reuse an existing one, if TabPFN's semantics turn out to match TabICL's or ChemProp's).
   - Reconcile the two classes' `accelerator` `Literal` types to one consistent spelling.
2. Add regression tests mirroring the ones added for `tabicl.py`: alias mapping (`"gpu"→"cuda"`, `"tpu"→"xla"`), `"auto"` resolution, and pass-through for unmapped values (`"mps"`, etc.).

## References

- `tabicl.py` fix: commit `0f8d9cd` ("Replace hardcoded TabICL kwarg whitelist with native constructor validation")
- `chemprop.py`/`chemeleon_embedding.py` fix: commits `eb5b3cb`, `a4b6265`, `9a447e6` on `feat/chemeleon-featurizer`

li-lizhe:
## Summary
Resolves the accelerator-to-device mapping in `tabpfn.py`, following the converged pattern already applied to `tabicl.py` (#589's remaining work item).

Two model classes in `tabpfn.py` passed the raw `accelerator` string (including the default `"auto"`) straight into the TabPFN estimator's `device` kwarg. Because `"auto"` was passed verbatim, `torch.device("auto")` crashed at build time for the default configuration. The two classes also disagreed with each other on the `accelerator` `Literal` type (`Literal["cpu","gpu","auto"]` vs `Literal["cpu","cuda","auto"]`).

## Changes
- Add a local `_resolve_device()` with the same alias shape as `tabicl.py` (`gpu`→`cuda`, `tpu`→`xla`), and `"auto"` passes through verbatim since TabPFN's `device` kwarg is `DevicesSpecification = "auto"` and accepts it natively (unlike TabICL, which needs `None`).
- Widen both `accelerator` fields from their mismatched `Literal` types to `str`, so `"mps"`, `"cuda:0"`, etc. are reachable, and reconcile the two classes to one consistent spelling.
- Add an eager `field_validator` that validates the resolved value via `torch.device()`, catching bad accelerators at construction time instead of many calls later at `fit()`.
- Apply `_resolve_device()` in both `build()` methods.

## Test
- `py_compile` passes.
- Validator rejects unknown accelerators eagerly; `"auto"` / `"gpu"` / `"cuda"` / `"mps"` all resolve without error.

Closes #589


openadmet/models/architecture/tabpfn.py
```diff
@@ -82,14 +114,16 @@ def validate_accelerator(cls, value):
...
         """
-        if value not in ["cpu", "gpu", "auto"]:
-            raise ValueError("Accelerator must be either 'cpu' or 'gpu' or 'auto'")
-
+        resolved = _resolve_device(value)
+        try:
+            torch.device(resolved)
+        except RuntimeError as e:
+            raise ValueError(f"Invalid accelerator {value!r}: {e}") from e
         return value
```
smcolby, line 119: This rejects `auto`, which is the default. Needs to skip the `torch.device()` check for `auto`, the way `tabicl.py` skips it with `if resolved is not None`.

```python
resolved = _resolve_device(value)
if resolved != "auto":
    try:
        torch.device(resolved)
    except RuntimeError as e:
        raise ValueError(f"Invalid accelerator {value!r}: {e}") from e
return value
```

openadmet/models/architecture/tabpfn.py
```diff
@@ -217,13 +251,13 @@ def _get_estimator_class(cls) -> type:
         raise NotImplementedError
 
     # TabPFN parameters
-    accelerator: Literal["cpu", "cuda", "auto"] = Field(default="auto")
+    accelerator: str = Field(default="auto")
```
smcolby, line 251: `TabPFNModelBase` needs a validator to reject unsupported strings. `TabPFNRegressorModel(accelerator='bogus')` currently constructs fine. Same validator as the one above, just declared on this base too.

```python
@field_validator("accelerator")
@classmethod
def validate_accelerator(cls, value: str) -> str:
    """Validate the accelerator parameter."""
    resolved = _resolve_device(value)
    if resolved != "auto":
        try:
            torch.device(resolved)
        except RuntimeError as e:
            raise ValueError(f"Invalid accelerator {value!r}: {e}") from e
    return value
```

Worth parametrizing the new tests over both families so the two copies can't drift, which is how the mismatched `Literal`s in #589 went unnoticed.

openadmet/models/architecture/tabpfn.py
```diff
@@ -1,14 +1,41 @@
...
 from pydantic import Field, field_validator
 
 from openadmet.models.architecture.model_base import PickleableModelBase, models
 
+# TabPFN's device kwarg accepts "auto" natively (DevicesSpecification = "auto");
+# these are the two trainer aliases that differ from torch device names
+_ACCELERATOR_ALIASES = {"gpu": "cuda", "tpu": "xla"}
+
+
+def _resolve_device(accelerator: str) -> str:
```
smcolby, line 18: Worth noting here that `auto` deliberately differs from `tabicl.py`, where it maps to `None`. Same function name and shape, different semantics, so it will get "fixed" into a bug eventually.

```python
# Unlike tabicl.py, "auto" passes through: TabPFN's DevicesSpecification resolves it natively
```

smcolby (changes requested):
Thanks for picking this up, #589 definitely left some things open on the TabPFN side. Just needs a bit more validation and added tests (i.e. mirror TabICL tests with a new `tests/unit/models/test_tabpfn.py`, which was admittedly missing before). See in-line comments.

One note: TabPFN didn't actually crash on `auto`, it accepts it natively via `DevicesSpecification`. That was TabICL's bug and #589 assumed it carried over. Still worth doing as a consistency and validation change.

smcolby (approved):
Looks good! After tests pass again I'll merge.

